PR #15, opened 2016-09-16
Reverse Sync

carolinux:


qfieldsync/dialogs/synchronize_dialog.py
```diff
@@ -0,0 +1,61 @@
...
+        self.push_btn.clicked.connect(self.start_synchronization)
+        self.button_box.addButton(self.push_btn, QDialogButtonBox.ActionRole)
+        self.qfieldDir.setText(plugin_instance.get_import_folder())
+        self.qfieldDir_btn.clicked.connect(make_folder_selector(self.qfieldDir))
+
+    def start_synchronization(self):
+        qfield_folder = self.qfieldDir.text()
+        qgs_file = get_children_with_extension(qfield_folder, 'qgs', count=1)[0]
+        open_project(qgs_file)
+        QgsOfflineEditing().synchronize() # no way to know exactly if it succeeded?
```
mbernasocchi: you could listen to https://qgis.org/api/classQgsOfflineEditing.html#a9b449781dd23b742616301d621b60688

qfieldsync/dialogs/synchronize_dialog.py
```diff
@@ -0,0 +1,61 @@
...
+from qfieldsync.utils.file_utils import get_children_with_extension
+from qfieldsync.utils.qgis_utils import open_project
+from qfieldsync.utils.qt_utils import get_ui_class, make_folder_selector
+from qfieldsync.config import *
+
+FORM_CLASS = get_ui_class('synchronize_base.ui')
+
+
+class SynchronizeDialog(QtGui.QDialog, FORM_CLASS):
+    def __init__(self, iface, plugin_instance):
```
mbernasocchi: maybe call this PullDialog since the other is a PushDialog

qfieldsync/dialogs/synchronize_dialog.py
```diff
@@ -0,0 +1,61 @@
...
+        self.iface = iface
+        self.push_btn = QPushButton(self.tr('Synchronize'))
+        self.push_btn.clicked.connect(self.start_synchronization)
+        self.button_box.addButton(self.push_btn, QDialogButtonBox.ActionRole)
+        self.qfieldDir.setText(plugin_instance.get_import_folder())
+        self.qfieldDir_btn.clicked.connect(make_folder_selector(self.qfieldDir))
+
+    def start_synchronization(self):
+        qfield_folder = self.qfieldDir.text()
+        qgs_file = get_children_with_extension(qfield_folder, 'qgs', count=1)[0]
```
mbernasocchi: refactor this to a comodity method get_project_in_folder

qfieldsync/dialogs/synchronize_dialog.py
```diff
@@ -0,0 +1,61 @@
...
+        open_project(qgs_file)
```
mbernasocchi: so the idea here is to open the project that was copied from the device to a location on the localhost and to close whatever is open currently.

I see some issues:
- the currently opened project is wiped away without prompting for saving.
- shouldnt we have the original (as in before pushing) project open and syncronise there?
- can you explain me your line of thoughts?

carolinux: 1) I assumed if the project is 'dirty' there would be a prompt from QGis but I can double check
2+3) I discussed this in person with @m-kuhn (I had some concerns along the same line as you) - what the goal of the Pull is (right now at least), is to synchronize the changes to the remote layers done in QField to the original postgres databases. We therefore don't care about the original project.

mbernasocchi: ok, so pleas act on point 1 an let 2 and 3 for future releases

mbernasocchi:
@carolinux I moved this code to the reverse sync branch

qfieldsync/dialogs/synchronize_dialog.py
mbernasocchi: done

qfieldsync/dialogs/synchronize_dialog.py
mbernasocchi: done

qfieldsync/dialogs/synchronize_dialog.py
mbernasocchi: done

PR #53, opened 2017-12-12
[fix #46] add option copy if missing

### LINKED ISSUE #46: Check if copy is needed for big files (e.g. basemaps)
It would be nice to have an option that checks whether layer that should be copied already exist (in the same state) at the destination path. Or maybe an option that does not actually copy specified layers but adds them in the .qgs-Project file as a layer that is present. 

I am thinking mainly about base maps that can be very large and take many minutes to copy. We used the sync plugin during field work every evening to set the project offline for next days work. The acutal data are very small in size, but the basemap always took very long to copy to the offline project. QFieldSync allows to "remove" some layers but then of course you also don't have them on QField.

3nids:
this keep existent data on the disk

### DIFF
--- a/qfieldsync/core/layer.py
+++ b/qfieldsync/core/layer.py
@@ -49,6 +49,9 @@ class SyncAction:
     # - the file(s) will be copied
     NO_ACTION = "no_action"
 
+    # Keep already copied data or files if existent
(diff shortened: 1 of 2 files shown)


3nids:
wait for PR #52 to get a green light I guess

PR #54, opened 2017-12-13
[fix #45] avoid duplicating composers when running export

### LINKED ISSUE #45: QGS project export duplicates composer
Hello,

I'm working on a .qgs project using geopackages for field work. When exporting the main QGS project to an offline project, the composers of the main QGS project are duplicated. I therefore get an extra composer for each time I export the project to an offline project.

Expected behavior: Export doesn't create duplicate composers in source project

QGIS version: 2.18.11
QField Sync 1.0.0

3nids:


qfieldsync/core/offline_converter.py
```diff
@@ -47,8 +47,9 @@ class OfflineConverter(QObject):
     task_progress_updated = pyqtSignal(int, int)
     total_progress_updated = pyqtSignal(int, int, str)
 
-    def __init__(self, project, export_folder, extent, offline_editing):
+    def __init__(self, iface, project, export_folder, extent, offline_editing):
```
m-kuhn: I always try to keep iface out as much as possible. Then it's also available to standalone apps.

m-kuhn: e.g. here there is already project, can we just call clear on it?

3nids: that would work in master, but in 2.18 QgsProject::clear() is not doing a lot of stuff (I believe it's because composers are not attached to the project).

m-kuhn: Ok, let's go with this for now. Might be too much to create a callback just for that. Let's deal with this if we actually run into issues.

PR #97, opened 2019-01-10
"copy" option (instead of "no action") for GPKG layer

signedav:
So the GPKG file is copied if a layer has this option.
In case there are several layer in this file, they are all copied. But only the ones with "copy" option are used from there. In case a GPKG layer has the option "offline editing" it has as the source the usual "data.gpkg" and if the GPKG layer has the option "remove" it's not in the project at all.

This fixes https://github.com/opengisch/Vogelwarte_QField_2019/issues/7

### DIFF
--- a/qfieldsync/core/layer.py
+++ b/qfieldsync/core/layer.py
@@ -95,7 +95,8 @@ class LayerSource(object):
 
     @property
     def is_file(self):
-        if os.path.isfile(self.layer.source()):
+        # reading the part before | so it's valid when gpkg
+        if os.path.isfile(self.layer.source().split('|')[0]):
             return True
         elif os.path.isfile(QgsDataSourceUri(self.layer.dataProvider().dataSourceUri()).database()):
             return True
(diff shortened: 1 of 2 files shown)


qfieldsync/core/layer.py
```diff
@@ -153,16 +154,16 @@ def copy(self, target_path):
...
+        file_path = self.layer.source().split('|')[0]
         # Spatialite have the path in the table part of the uri
         uri = QgsDataSourceUri(self.layer.dataProvider().dataSourceUri())
 
         if os.path.isfile(file_path):
             source_path, file_name = os.path.split(file_path)
             basename, extensions = get_file_extension_group(file_name)
             for ext in extensions:
-                if os.path.exists(os.path.join(source_path, basename + ext)):
+                if os.path.exists(os.path.join(source_path, basename + ext)) and not os.path.exists(os.path.join(target_path, basename + ext)):
```
m-kuhn: With this change it's no longer possible to export into the same directory than a previous export and overwrite data files.
I'm not sure if that is an issue somewhere, but it may be better to keep track of already copied layers by passing a list of already copied files into this method and check against that list

PR #99, opened 2019-01-14
Only synchronize data-file once

signedav:
Checksum of last imported data.gpkg file (md5 hash) is stored in the project.
If trying to synchronize data from the same file (same checksum) again, it won't work.

### DIFF
--- a/qfieldsync/utils/file_utils.py
+++ b/qfieldsync/utils/file_utils.py
@@ -86,4 +86,5 @@ def import_file_checksum(folder):
     path = os.path.join(folder, "data.gpkg")
     if not os.path.exists(path):
         path = os.path.join(folder, "data.sqlite")
-    return hashlib.md5().hexdigest()
+    file_data = open(path, 'rb').read()
+    return hashlib.md5(file_data).hexdigest()
--- a/qfieldsync/utils/file_utils.py
+++ b/qfieldsync/utils/file_utils.py
@@ -22,6 +22,7 @@
 import os
 import platform
 import subprocess
+import hashlib
 
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


signedav:
Btw. it's not working correctly at the moment. The calculated checksum is mostly the same.

qfieldsync/utils/file_utils.py
```diff
@@ -79,3 +80,11 @@ def open_folder(path):
         subprocess.Popen(["open", '-R', path])
     else:
         subprocess.Popen(["xdg-open", path])
+
+
+def import_file_checksum(folder):
+    path = os.path.join(folder, "data.gpkg")
+    if not os.path.exists(path):
+        path = os.path.join(folder, "data.sqlite")
+    file_data = open(path, 'rb').read()
```
m-kuhn: This will produce a `ResourceWarning` no?

Better safe than sorry:

```suggestion
    with open(path, 'rb') as f:
        file_data = f.read()
```

m-kuhn:
Good job.
Did you consider to add a list of imported hashes instead of only the last one? It's not a customer requirement, but might safe someone some troubles in the future.

signedav:
Would make sense to have a list I guess. Can do the day after tomorrow I guess.

m-kuhn:
I think it's to indicate to other writing processes, that there's a reading process, but I'm not sure by any means.
If you need more reliable information, we need to ping Even.

signedav:
I just wondered because I've seen that after an `open_project(qgs_file)` the `data.gpkg` is not the same anymore. Means new edit-date and of course new MD5 checksum... But since I calculate the checksum now before the first `open_project(qgs_file)` this works. 

Just be aware that if I synchronize the project - then I open the qfield-project again manually or whatever but don't change anything -> the `data.gpkg` is updated and the MD5 hash won't be the same. But I guess this is okay.

PR #126, opened 2019-10-14
update file suffixes for shape, mapinfo tab, png, jpg and tif

damstep:


qfieldsync/core/layer.py
```diff
@@ -13,10 +13,9 @@
 # When copying files, if any of the extension in any of the groups is found,
 # other files with the same extension in the same folder will be copied as well.
 file_extension_groups = [
-    ['.shp', '.shx', '.dbf', '.sbx', '.sbn', '.shp.xml', '.qix']
+    ['.shp', '.shx', '.dbf', '.sbx', '.sbn', '.shp.xml', '.prj','.cpg','.qpj','.qix']['.tab','.dat','.map','.xls','.xlsx','.id','.ind','.wks','.dbf']['.png','.pgw']['.jpg','.jgw']['.tif','.tfw']
```
m-kuhn:
```suggestion
    ['.shp', '.shx', '.dbf', '.sbx', '.sbn', '.shp.xml', '.prj','.cpg','.qpj','.qix'],
    ['.tab','.dat','.map','.xls','.xlsx','.id','.ind','.wks','.dbf'],
    ['.png','.pgw'],
    ['.jpg','.jgw'],
    ['.tif','.tfw']
```

m-kuhn:
Thanks a lot, this looks much better!

I have left a suggestion (I guess a couple of commas are missing), can you accept them?
Then I would suggest you download this modified file from github and replace it locally inside your plugin folder and check if the patch actually works. (I can help you find the download link for the file after you have accepted the suggestion)

damstep:
ok have accepted, just cant find the download for the corrected version

damstep:
dont woryy found it.

damstep:
tested on shape and png files and all subsidiary files come across to the new directory. plus the qgs project opens fine.
looks good.

m-kuhn:
Thanks!

3nids:
I believe it misses *.wld extension for the images format at least

m-kuhn:
@damstep would you mind adding this one too in a new pull request? Just like the current one.

damstep:
can do as a temp measure. Just wondering if it might make more sense just to copy everything in the folder with the same name regardless of suffix? would be simpler and likely wouldn't grab too much extraneous data.
Also ermapper files use a file with no suffix, which i cant figure out how to add using this code ie:
someraster.ers
someraster
someraster.ecw
the file with no suffix actually contins all of the data.
thanks

PR #133, opened 2020-01-17
add table to define photo naming

3nids:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -164,5 +164,6 @@ fabric.properties
 # My super new category
 .DS_Store
 
-resources_rc.py
+qfield-sync.*.zip
+qfieldsync/resources_rc.py
 
--- a/qfieldsync/utils/exceptions.py
+++ b/qfieldsync/utils/exceptions.py
@@ -19,17 +19,12 @@
  ***************************************************************************/
 """
 
-# coding=utf-8
-from qgis.core import (
-        QgsMessageLog,
-        Qgis
-)
+from qgis.core import QgsMessageLog, Qgis
 
 
 class QFieldSyncError(Exception):
 
-    def __init__(
-            self, message, exception=None, long_message=None, tag="QFieldSync"):
+    def __init__(self, message, exception=None, long_message=None, tag="QFieldSync"):
         """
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -164,5 +164,6 @@ fabric.properties
 # My super new category
 .DS_Store
 
-resources_rc.py
+qfield-sync.*.zip
```
m-kuhn: Is the `-` (dash) on purpose?

qfieldsync/utils/exceptions.py
```diff
@@ -19,17 +19,12 @@
  ***************************************************************************/
 """
 
-# coding=utf-8
-from qgis.core import (
-        QgsMessageLog,
-        Qgis
-)
+from qgis.core import QgsMessageLog, Qgis
```
m-kuhn: The idea of the multiline import is to keep them trackable in git

.gitignore
3nids: yep

qfieldsync/utils/exceptions.py
3nids: the IDE makes it clear which is used or not.
is there another purpose?

m-kuhn: It's about git history, not the ide

3nids: I just don't see the interest in tracking import changes and I prefer the cleaner code layout.
But I don't mind reverting, just tell me

m-kuhn: I'd like to have a common style here, I was always switching things TO this style, but switching forth and back seems counterproductive.

We could also go for

```
from x import Y
from x import Z
```

that would also avoid extremely long lines

signedav:
I had to create the resources.py manually.  
`pyrcc5 resources.qrc -o resources_rc.py`
Is there missing something or is it just me?

m-kuhn:
You could have used `make`. In the old days we used to "compile" images and dialogs into resource py files, as we also do with cpp projects still often. This defeats one of the nice things of python (no compilation involved) so we no longer do this a lot nowadays.

PR #135, opened 2020-02-25
save last used folder for import/export as a project property

### LINKED ISSUE #124: Use unmodified filepath from preferences
In the preferences of the qfieldsync plugin you can define where the offline files should be saved to. When you actually use it, the file path is always extended by the QGIS-Project name. We find this not useful and propose to export files directly to the export directory (with out adding a folder).

Filepaths I would want:
grafik

Filepath qfieldsync proposes (added the "red kite" with is the QGIS project name):
grafik

3nids:
fixes #124


3nids:
Tests must be updated since the signatures of the dialogs changed.
Otherwise, should be good to go.

m-kuhn:
@3nids when are the tests planned?

m-kuhn:
Thanks !

PR #136, opened 2020-03-02
Photo_path fixes

signedav:
- Check if layer is vector layer on attachment fields
- use row as index

Before I received this:
```
2020-03-02T14:48:25     WARNING    Traceback (most recent call last):
              File "/home/david/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/package_dialog.py", line 94, in 
              self.infoLabel.linkActivated.connect(lambda: self.show_settings())
              File "/home/david/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/package_dialog.py", line 153, in show_settings
              dlg = ProjectConfigurationDialog(self.iface, self.iface.mainWindow())
              File "/home/david/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/project_configuration_dialog.py", line 79, in __init__
              self.reloadProject()
              File "/home/david/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/project_configuration_dialog.py", line 120, in reloadProject
              for field in layer.fields():
             AttributeError: 'QgsRasterLayer' object has no attribute 'fields'
```
And I got empty fields...

### DIFF
--- a/qfieldsync/gui/project_configuration_dialog.py
+++ b/qfieldsync/gui/project_configuration_dialog.py
@@ -126,14 +126,14 @@ class ProjectConfigurationDialog(QDialog, DialogUi):
                         item = QTableWidgetItem(layer.name())
                         layer_source = LayerSource(layer)
                         item.setData(Qt.UserRole, layer_source)
-                        self.photoResourceTable.setItem(count, 0, item)
+                        self.photoResourceTable.setItem(row, 0, item)
                         item = QTableWidgetItem(field.name())
-                        self.photoResourceTable.setItem(count, 1, item)
+                        self.photoResourceTable.setItem(row, 1, item)
                         ew = QgsFieldExpressionWidget()
                         ew.setLayer(layer)
                         expression = layer_source.photo_naming(field.name())
                         ew.setExpression(expression)
-                        self.photoResourceTable.setCellWidget(count, 2, ew)
+                        self.photoResourceTable.setCellWidget(row, 2, ew)
                         row += 1
         self.photoResourceTable.resizeColumnsToContents()
 
--- a/qfieldsync/gui/project_configuration_dialog.py
+++ b/qfieldsync/gui/project_configuration_dialog.py
@@ -24,7 +24,7 @@ from qgis.PyQt.QtGui import QIcon
 from qgis.PyQt.QtWidgets import QDialog, QTableWidgetItem, QToolButton, QComboBox, QMenu, QAction
 from qgis.PyQt.uic import loadUiType
 
-from qgis.core import QgsProject, QgsMapLayerProxyModel
+from qgis.core import QgsProject, QgsMapLayerProxyModel, QgsMapLayer
 from qgis.gui import QgsFieldExpressionWidget
 
 from qfieldsync.core import ProjectConfiguration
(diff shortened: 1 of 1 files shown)


signedav:
@3nids Additionally to these fixes I would like to discuss these things:
- As far as I understand that on apply it adds the custom properties to the original project. The offline project never gets them. Is something missing?
- I'm not sure if you have been in this discussion on a weekly teammeeting. We talked about the sub-path, that this should be defined as well. Means it would be nice to have as a default value `DCIM/` included.
- Should the extension (.jpg) be added by QField? If so we would need to make a note somewhere in the GUI.

3nids:
Thanks for the fix!

> As far as I understand that on apply it adds the custom properties to the original project. The offline project never gets them. Is something missing?

It might be an issue in offline editing. I will have a look.

> I'm not sure if you have been in this discussion on a weekly teammeeting. We talked about the sub-path, that this should be defined as well. Means it would be nice to have as a default value DCIM/ included.

Is this what you mean https://github.com/opengisch/qfieldsync/pull/137 ?

> Should the extension (.jpg) be added by QField? If so we would need to make a note somewhere in the GUI.

I would say that we test if `\.jpe?g/i` is the exstension and we append if not. Works for you?

signedav:
> I would say that we test if \.jpe?g/i is the exstension and we append if not. Works for you?

Okaydokey.

m-kuhn:
No risk that someone uses .png, .raw., .JPEG, ...?

3nids:
> .png, .raw.

well, if then, jpg is going to be appended

>  .JPEG

this is handled by the nice regex ;)

signedav:
As far as I understand our existing code, the photos taken by the camera are stored as a `.jpg`.

m-kuhn:
> > .png, .raw.
> 
> well, if then, jpg is going to be appended
> 

which would render the file useless, which is my concern



> As far as I understand our existing code, the photos taken by the camera are stored as a `.jpg`.

Native one also? If that's forced anyway, no further questions.

signedav:
I see three possibilities:
- we have no extension in our default expression - means we make the user to not define an expression -> and then we add jpg like it's done currently.
- we parse the file for an extension. Whatever it is, we put the camera result in there. Leaf it to the user, if he tries to store it wherever. \*
- we parse for jpg / jpeg like @3nids suggested...

 \* this could maybe be reused for upcoming attachments with other file types.

3nids:
> which would render the file useless, which is my concern

why? are there some system camera which produces other files than JPG?
if so, we could take the other approach: strip away any extension from the expression and append the extension from the system file name?

signedav:
WhatsApp Image 2020-03-04 at 10 03 41
Open Camera App

So I suggest:
- QFieldSync has the default expression without any extension (maybe tool tip)
- Remove extension
- If I can get the type from the camera, I use it
- Otherwise append ".jpg"

Fine?

3nids:
perfect to me

signedav:
I committed the light version with replacing any extension with`.jpg`

Why I did not detect the type returned by the camera? Because the camera does not return any type, not even a file name. Currently it's solved by passing the camera the filename (and path) where to store the image. This means currently only JPGs are supported.

PR #142, opened 2020-03-30
Synchronize files of DCIM folder

signedav:
On import copy files from DCIM folder to original project path and on export to offline project path.

Subfolders are created if they don't exist.
Existing files are overwritten.

Resolves ##93


qfieldsync/utils/file_utils.py
```diff
@@ -104,3 +105,22 @@ def slugify(text: str) -> str:
     slug = re.sub(r'[-]+', '-', slug)
     slug = slug.lower()
     return slug
+
+
+def copy_images(source_folder, destination_folder):
+    if os.path.isdir(source_folder):
+        if not os.path.isdir(destination_folder):
+            os.mkdir(destination_folder)
```
m-kuhn: Can we always assume the parent folder exists in every case?
I guess yes, but if not https://stackoverflow.com/a/600612/2319028

signedav: In the context, yes. Since we only call this, when there is a project-folder.

signedav:
@m-kuhn I inserted tests in the export test.
The import test is not yet implemented. Should I make a general import test in general or is it sufficient like this?

signedav:
Okay to merge @m-kuhn 
We can improve the general export tests another time as well.

PR #148, opened 2020-05-27
Make changelog only for current release

signedav:


m-kuhn:
Makes sense to me.
Should we add a link to https://github.com/opengisch/qfieldsync/releases where they can see images and older changelog entries too? The current visualization in the qgis plugin manager is kind of limited.

We could also add it as a todo for plugin_ci to generate that automatically on release.

suricactus:
It is very common that users jumps to a new version skipping a few inbetween. Sometimes skipping even a few major releases, the average Joe hates updates. I think a link to https://github.com/opengisch/qfieldsync/releases is a must and can avoid a lot of headache for both sides.

signedav:
image
Sadly html is not allowed in this section to create a link. So I hope this is sufficient.

m-kuhn:
@signedav do you have an idea why the test is failing?

signedav:
Checking right now. Weird that they did not failed when the DCIM synchronization has been integrated...

PR #151, opened 2020-06-01
Automatically add related layer from a value relation widget config

3nids:
and report error if it missing

When setting a layer as being used, if it has dependent layers, they will be automatically used too (in the project config).
In the package layer dialog, error is reported if dependent layers are missing. 

For now, the dependent layers are checked for value relation widget config.
This can be easily extended for other use cases.

ezgif-1-ee2fc731f4e4

### DIFF
--- a/qfieldsync/core/configuration_checkup.py
+++ b/qfieldsync/core/configuration_checkup.py
@@ -1,3 +1,23 @@
+# -*- coding: utf-8 -*-
+
+"""
+/***************************************************************************
+ QFieldSync
+                              -------------------
+        begin                : 2020
+        copyright            : (C) 2020 by OPENGIS.ch
+        email                : info@opengis.ch
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+ """
 
 from qgis.PyQt.QtCore import QObject
 from qgis.core import QgsProject, Qgis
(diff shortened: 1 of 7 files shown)


qfieldsync/ui/package_dialog.ui
```diff
@@ -240,10 +240,13 @@
       <item row="0" column="0">
        <widget class="QLabel" name="infoLabel">
         <property name="text">
-         <string>Some layers in this project have not yet been configured. &lt;a href=&quot;configuration&quot;&gt;Configure project now&lt;/a&gt;.</string>
+         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Some issues has been found. &lt;a href=&quot;configuration&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#419cff;&quot;&gt;Configure project now&lt;/span&gt;&lt;/a&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
m-kuhn, line 243:
```suggestion
         <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;Some issues have been found. &lt;a href=&quot;configuration&quot;&gt;&lt;span style=&quot; text-decoration: underline; color:#419cff;&quot;&gt;Configure project now&lt;/span&gt;&lt;/a&gt;.&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```

qfieldsync/core/configuration_checkup.py
```diff
@@ -0,0 +1,39 @@
+
```
m-kuhn, line 1: License header

m-kuhn:
@3nids 

What do you think about this?

> I think this is always shown when first setting up a project. Do we want to tell the user that "issues have been found"?

3nids:
we can reformulate to something closer than the current formulation, or you would prefer to say nothing?

m-kuhn:
Maybe we can show different messages on "issues" or "no configuration done yet"?

m-kuhn:
@3nids do we continue with that?

suricactus:
Nice feature. :+1: Too many conflicts. :-1:  Closing. :handshake:

PR #157, opened 2020-06-15
FEATURE add option to lock geometries

suricactus:
Add a new layer option to "lock geometries". This new option should ideally live on a new vector layer properties panel. QField should respect this new configuration option and not allow adding, splitting, modifying etc. geometries.

Peek 2020-06-17 12-48

https://github.com/opengisch/QField/pull/1063 functionality:
Peek 2020-06-15 11-32

### DIFF
--- a/qfieldsync/ui/map_layer_config_widget.ui
+++ b/qfieldsync/ui/map_layer_config_widget.ui
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>QFieldLayerSettingsPage</class>
+ <widget class="QWidget" name="QFieldLayerSettingsPage">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>500</width>
+    <height>250</height>
+   </rect>
+  </property>
+  <property name="windowTitle">
+   <string>Form</string>
+  </property>
+  <layout class="QFormLayout" name="QFieldLayerSettingsPageLayout">
+   <item row="0" column="0">
+    <widget class="QLabel" name="layerActionLabel">
+     <property name="text">
+      <string>Layer Action</string>
+     </property>
+    </widget>
+   </item>
+   <item row="0" column="1">
+    <widget class="QComboBox" name="layerActionComboBox">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="1">
+    <widget class="QCheckBox" name="isGeometryLockedCheckBox">
+     <property name="toolTip">
+      <string>When enabled, this option disables adding and deleting features, as well as modifying the geometries of existing features.</string>
+     </property>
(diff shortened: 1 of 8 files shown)


qfieldsync/utils/qt_utils.py
```diff
@@ -29,3 +29,18 @@ def selectFolder(line_edit_widget):
 
 def make_folder_selector(widget):
     return partial(selectFolder, line_edit_widget=widget)
+
+
+def set_available_actions(combobox, layer_source):
```
3nids, line 34: that doesn't seem to belong to qt_utils. You can rather create a gui/utils.py ?

suricactus: TBH I had no idea where to put it. The function gets `QComboBox` as a param, similar to `def selectFolder(line_edit_widget: QLineEdit)` in the same file, that's why I thought is a good place to go.

suricactus:
Sure, I will add photo naming config too. Thanks for the suggestion.

qfieldsync/utils/qt_utils.py
suricactus: Should I really move it to another file, tried it and it feels like one more file to manage, rather than clear separation of the functionality? No strong opinions, just slight feelings.

3nids: it's just that it contains QFieldSync specific code and could not be reused in another plugin, while the rest of the file yes.
I'd prefer a new file personally, but no biggie.

suricactus: Deal

suricactus:
@3nids I think I am done with the photo naming configuration, you can have a look.

qfieldsync/ui/map_layer_config_widget.ui
```diff
@@ -0,0 +1,48 @@
...
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="1">
+    <widget class="QCheckBox" name="isGeometryLockedCheckBox">
+     <property name="toolTip">
+      <string>When enabled, this option disables adding and deleting features, as well as modifying the geometries of existing features.</string>
+     </property>
+     <property name="text">
+      <string>lock geometry</string>
```
m-kuhn, line 40:
```suggestion
      <string>Lock Geometries</string>
```

Just the usual nitpicking

suricactus: NOTE that not all of the checkboxes are in This Style, e.g. `Create base map`

m-kuhn: https://docs.qgis.org/3.10/en/docs/developers_guide/hig.html

For what I can see, rule 3. applies here.

suricactus: Then as far as I understand it, it should be `Create Base Map`, we have a selectable widget `QCheckBox` and there are no prepositions, conjuctions and articles that are the only exceptions? 

I don't know, I just checked other QGIS interfaces, it seems I am not the only one who got confused by rule 3 :)

m-kuhn: Yes, that's what it should be. QGIS has organically grown and it's a movement towards Zen

suricactus: I will make a separate PR on this later, there are few more captions that don't follow the rule

PR #162, opened 2020-07-07
Use three dots instead of elipsis

suricactus:
pylupdate5 hardcodes the output to ascii symbols, therefore any unicode is XML encoded and cannot be reverted to the original state.

Enforcing `codecForTr` to be utf8 does not work:
https://github.com/baoboa/pyqt5/blob/31ef42563dbcaf3885da8803e2e5d84507141c9b/pylupdate/pylupdate_main.py#L231
As the output is hardcoded as ASCII in
https://github.com/baoboa/pyqt5/blob/31ef42563dbcaf3885da8803e2e5d84507141c9b/pylupdate/metatranslator.cpp#L435

### DIFF
--- a/qfieldsync/core/offline_converter.py
+++ b/qfieldsync/core/offline_converter.py
@@ -103,7 +103,7 @@ class OfflineConverter(QObject):
             for layer in self.__layers:
                 original_layer_info[layer.id()] = (layer.source(), layer.name())
 
-            self.total_progress_updated.emit(0, 1, self.tr('Creating base map...'))
+            self.total_progress_updated.emit(0, 1, self.trUtf8('Creating base map…'))
             # Create the base map before layers are removed
             if self.project_configuration.create_base_map:
                 if 'processing' not in qgis.utils.plugins:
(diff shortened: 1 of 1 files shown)


m-kuhn:
:cry:

m-kuhn:
image

does this help us to recover utf-8 from those mangled poor characters? Will need a bit of preprocessing of the string though, but having proper unicode support in qgispluginci would be great.

suricactus:
Unfortunately not, because the characters are encoded byte by byte, not char by char. 
```
>>> import html
>>> html.unescape('&#1086;&#1087;&#1072;') # cyrillic "опа" (latin: opa)
'опа'
>>> html.unescape('&#xe2;&#x80;&#xa6;') # elipsis is encoded for each individual byte
'â€¦'
```

m-kuhn:
what about creating a .pro?

suricactus:
It is using the same `codecForTr` parameter as in the first post, it does not work.

m-kuhn:
How far did you go with migration of pylupdate to qgis-plugin-ci?

suricactus:
I am testing https://github.com/opengisch/qgis-plugin-ci/pull/18 and if all good, then I will revert the changes here.

suricactus:
Just to unblock the change to go further, temporary use the mask as obsolete method `trUtf8` instead of `tr`. There is no other disadvantage of using the `trUtf8` for now, except a warning and possible post-qt5 era need for porting.

https://doc.qt.io/qt-5/qobject-obsolete.html#trUtf8

When the `.pro` workaround is proved to work properly and merged at qgis-plugin-ci, we can get rid of `trUtf8` calls.

PR #171, opened 2020-08-20
Localized data path awareness and handling

nirvn:
@m-kuhn , as discussed over whatsapp, I took the time to update the QFieldSync plugin to be aware of and handle localized data path stored layers.

What the PR does:
- when layer(s) is/are detected as stored in a localized data path, the user is informed and reminded of the need for those to be copied in devices of QField users
- layers stored in a localized data path are not copied (and the list of possible actions for those layers updated accordingly)

While at it, I've updated the plugin to make use of the decodeUri and encodeUri functions, which allows the plugin to properly copy files by non OGR/GDAL providers. For e.g., delimited text provider.

### DIFF
--- a/qfieldsync/core/layer.py
+++ b/qfieldsync/core/layer.py
@@ -7,7 +7,10 @@ from qgis.PyQt.QtCore import QCoreApplication
 from qgis.core import (
     QgsDataSourceUri,
     QgsMapLayer,
-    QgsReadWriteContext
+    QgsReadWriteContext,
+    QgsProject,
+    QgsProviderRegistry,
+    QgsProviderMetadata
 )
 
 from qfieldsync.utils.file_utils import slugify
@@ -74,6 +77,17 @@ class LayerSource(object):
         self._is_geometry_locked = None
         self.read_layer()
 
+        self.storedInlocalizedDataPath = False
+        if layer.dataProvider() is not None:
+            pathResolver = QgsProject.instance().pathResolver()
+            md = QgsProviderRegistry.instance().providerMetadata(layer.dataProvider().name())
+            if md is not None:
+                decoded = md.decodeUri(layer.source())
+                if "path" in decoded:
+                    path = pathResolver.writePath(decoded["path"])
+                    if path.startswith("localized:"):
+                        self.storedInlocalizedDataPath = True
+
     def read_layer(self):
         self._action = self.layer.customProperty('QFieldSync/action')
         self._photo_naming = json.loads(self.layer.customProperty('QFieldSync/photo_naming') or '{}')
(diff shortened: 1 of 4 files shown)


nirvn:
Hm, so decodeUri/encodeUri are not in RIP 3.4, but are present in current 3.10 LTR. I didn't realize you guys had a 3.4 check in there. Are you deliberately maintaining support for 3.4 or it's just an old travis test which should be updated to 3.10? :)

m-kuhn:
update to 3.10 totally doable !

m-kuhn:
but let's reflect that in metadata.txt

nirvn:
Worth it. I can handle the metadata.xml, can you update travis?

m-kuhn:
It's even newer than 3.10. Can we trigger the functionality only if QGIS is newer than 3.12 ?

nirvn:
Ah, encodeUri is 3.12. I'll adjust.

nirvn:
@suricactus , gentle ping :) I'd love for a quick PR review so we can merge this.

As I mentioned above, on top of adding localized data path handling, I applied some UI/UX improvements along the way. 

The package dialog has gone better:
- a spacer was added to avoid overly spacy widgets when vertically expanding the dialog
- a configure project button has been added at the bottom of the dialog and is _always_ visible (instead of the information hyperlink)
- Instead of using a custom 'Create' button, we're using the dialog button box's Save button (better placement, system icon, etc.)
Screenshot from 2020-08-21 10-01-59

The synchronize package also got some improvements (essentially porting the relevant package dialog bits):
Screenshot from 2020-08-21 10-02-08

Finally, QGIS' message bar has been handling folder/file hyperlinks for a while now, so there is no need for qfieldsync to push a widget. Instead, merely push a string and let QGIS handle everything else. The resulting message bar looks a lot better (margin/padding):
Screenshot from 2020-08-21 10-01-52

qfieldsync/core/layer.py
```diff
@@ -74,6 +78,17 @@ def __init__(self, layer):
         self._is_geometry_locked = None
         self.read_layer()
 
+        self.storedInlocalizedDataPath = False
+        if self.layer.dataProvider() is not None:
+            pathResolver = QgsProject.instance().pathResolver()
+            md = QgsProviderRegistry.instance().providerMetadata(self.layer.dataProvider().name())
```
suricactus, line 84: not a strong opinion here, but I would personally call all these variables `meta` or `metadata`, instead of `md`.

m-kuhn: Couldn't we call it "markdown" ? :stuck_out_tongue_winking_eye:

qfieldsync/ui/package_dialog.ui
```diff
@@ -238,9 +254,22 @@
...
+        </property>
+        <property name="wordWrap">
+         <bool>true</bool>
+        </property>
+       </widget>
+      </item>
+      <item row="1" column="0">
+       <widget class="QLabel" name="infoLocalizedPresentLabel">
+        <property name="text">
+         <string>Some layers are stored in localized data paths, make sure those are copied into the localized data path of devices running QField. On most devices, the path is /QField/basemaps.</string>
```
m-kuhn, line 269:
```suggestion
         <string>Some layers are stored in localized data paths, make sure those are copied into the localized data path of devices running QField. On most devices, the path is <pre>/QField/basemaps.</pre></string>
```

That will make it clear this belongs together (compare the break placed in a strange position in your screenshots)

m-kuhn: What would be very nice (easy to do?): print a list of layer /filenames.

PR #175, opened 2020-08-23
Remove individual changelog entries

m-kuhn:
Let's point only to things we are sure are up to date

Thanks for the suggestions @nirvn


signedav:
Can we close that @m-kuhn or should it change the string from here f3c2fef3a4a62fceb43a044cc4985893784aac47 ?

m-kuhn:
:broom:

PR #177, opened 2020-08-27
QFieldCloud integration

suricactus:
Allows projects to be uploaded to QFieldCloud via this plugin.

This is work in progress, but general overview of the available UI is there. I will regularly update the screenshots if needed.

### Login

First run of the plugin in the current session asks for credentials (unless they are already remembered):

image

### Project management

Then a list of projects are loaded. On top right above the projects one can either create a new one, or refresh the list. The list is automatically refreshed after one creates, deletes, updates or opens the project data.

image

#### Update 
Projects can be updated. The local directory button on the right has a little drop down "use current project dir".
image

An opened project has two more tabs. The meta shows ... meta, and files, which lists all the files and their versions available on the cloud. The little download button on the right allows downloading a specific file version at specific location.

| meta | files |
|--------|-------|
| image | image |

#### Create

Creating a project is very much like editing one, except some of the fields are disabled. If the current project is already a cloud project, the "Local Directory" fielld is disabled:

image

#### Delete

One can delete a project, it asks for confirmation:

image

### Synchronize

When the sync button is clicked, there are several scenarios:

#### The project has no "Local Directory" set and has no files on the cloud

The user must select a local directory path, where might be a single `.qgs` project file.

#### The project has no "Local Directory" set and has some files on the cloud

The user must select an empty local directory path.

#### The project has already set "Local Directory"

This is the normal case. The user sees a file selection dialog with all the files that differ (sha256 checksum check), where he will chose which file to be overwritten - the one locally or the one on the cloud. If none of the checkboxes is selected, then the files will remain unsynced.

image

Once  the file section is done, the user press "Sync!" and all the files that will be overwritten by remote files are copied in `<project>/.qfieldsync/backup`. The files that are going to be uploaded are copied to `<project>/.qfieldsync/upload` and start uploading them. Once ready, it will download all the needed files from the cloud into `<project>/.qfieldsync/download`. If all is done, the files from download are copied over the current project files and the project is reloaded. If any of the steps fails, the files from `<project>/.qfieldsync/backup` are used to restore the initial state of the project and it is reloaded.

image

## Other

This plugin adds a few other UI elements:

### Browser tree item

To easily access all the projects.

image

### Toolbar button

Single click on this button makes a sync.

image

### Plugins -> QFieldSync menu items

image


suricactus:
Ideas:
- in Layer Properties -> QField (tab) one can browse the versions of the given layer (if file based)
- make sure multifile formats (shapefile, gpkg WAL etc) are always synced all together

nirvn:
@suricactus , branch rebased against master. Proof of life :wink: :
image

FYI, there's a nice sweet placeholder for QFieldCloud-related settings in the QField settings panel now:
image

You should consider adding the three QFieldCloud settings you added in preference.py in there.

Cheers

suricactus:
Added option to choose the server the user is connecting to:
image

suricactus:
TODO Should hide the `fid` of exported layers for offline editing.

lucienicolier:
When you sync, and go back to "synchronize with QFieldCloud" the files steel there. You need to refresh for them to disappear 
sync_cloud

lucienicolier:
every time I delete a project and then want to add a new one it's not possible
delete

suricactus:
If you logout/login from another device, the current session gets corrupted. Should logout instead.

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, network_manager: CloudNetworkAccessManager, parent: QWidget =
         else:
             self.welcomeLabelValue.setText(self.network_manager.username)
 
+        self.welcomeLabelValue.setToolTip(self.tr(f'You are logged in with the following username at {self.network_manager.url}'))
```
m-kuhn, line 103: Can we do this conditional (only show it when not default)?

suricactus: Done.

suricactus:
> every time I delete a project and then want to add a new one it's not possible

very weird problem, and very weird solution:

```diff
- self.createButton.clicked.connect(self.on_create_button_clicked)
+ self.createButton.clicked.connect(lambda: self.on_create_button_clicked())
```
The above is enough to fix all the disfunct buttons, not only the create one.

suricactus:
@lucienicolier your bug reports were addressed.

suricactus:
The sync dialog is a bit better, now there is the "detail" column, that explains what is going to happen. Anyway, open for discussion. Especially in file deletion, there are a number of undefined situation, unless I keep very complicated state in QFieldSync.

image

suricactus:
1) probably we need to change "No Action" layer action to "No Action (Keep as it is)".
2) Validate the project name to be without whitespaces.
3) on CloudTransferDialog we can use `QTreeView::resizeColumnToContents` since the first column is way too big.
4) Separate the two export option in the plugin menu.
5) When creating a new project, show a synchronize dialog.
6) When creating/updating a project , change the "submit" button to "create project"/"update information"
7) Geopackages needs to be flushed before checking for sync.
8) Delete project should be more red, last button in the overview and/or move to project details "danger zone".
9) Browser panel - double click to open. Sync does not work.
10) Show indication on the projects overview whether the project has a local checkout or not (probaly also show the path?).
11) Make sure the sync for projects that are not currently open works perfectly or disable it.
12) Improve the UI on sync screen with additional column between cloud/local, showing the direction of the transfer. 
13) Try to synchronize a project different from the current one, kaboom.

m-kuhn:
No action (directly access data source) ?

m-kuhn:
Or completely drop the "no action" and use only "directly access database".

suricactus:
Awesome improvement in the sync dialog from @olivierdalang:


Than you!

olivierdalang:
@suricactus One new issue it seems the "open project" is always disabled with `Cannot open project since no local .qgs or .qgz project file found` even when there is a local .qgz.

qfieldsync/core/cloud_api.py
```diff
@@ -57,7 +57,26 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+
+        try:
+            resp = json.loads(payload)
+            if resp.get('code'):
+                message = f'[{resp["code"]}] {resp["message"]}'
+            else:
+                message = resp['detail']
+        except:
+            if payload:
+                message = payload[:500] + '...'
```
m-kuhn, line 72:
```suggestion
                message = payload[:500]
                if len(payload) > 500:
                    message += '...'
```

suricactus: whoops, thanks

m-kuhn: Most important commit today !!

qfieldsync/core/cloud_api.py
```diff
@@ -69,7 +69,10 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
                 message = resp['detail']
         except:
             if payload:
-                message = payload[:500] + '...'
+                message = payload[:500] 
+
+                if len(payload) > 500:
+                    message += '…'
```
m-kuhn, line 75: Sorry for the missing ellipsis

suricactus: do i miss something here?

m-kuhn: I think you corrected my three points to a proper ellipsis

suricactus: Yup, I did a separate commit with real elllipsis, instead of three dots.

suricactus:
Find the recent changes in this range:
https://github.com/opengisch/qfieldsync/pull/177/files/e909e692edd600f71c896049e2811bf31a5abb66..2fd31ac

suricactus:
- [ ] Add longer timeouts for requests, currently cancelled around 60 seconds 
- [ ] Prevent uploading multiple `.qgs` or `.qgz` files

nirvn:
Fix added to prevent failure when downloading a project from the cloud containing attachment(s) with a path that hasn't been created locally yet.

PR #187, opened 2020-09-18
Fix error stack thrown when trying to import/synchronize data from a non-existing directory

nirvn:
Fix error stack thrown when trying to import/synchronize data from a non-existing directory

### DIFF
--- a/qfieldsync/utils/file_utils.py
+++ b/qfieldsync/utils/file_utils.py
@@ -44,6 +44,10 @@ def fileparts(fn, extension_dot=True):
 
 
 def get_children_with_extension(parent, specified_ext, count=1):
+    if not os.path.isdir(parent):
+        raise QFieldSyncError(
+            "The directory {} could not be found".format(parent))
+
     res = []
     extension_dot = specified_ext.startswith(".")
     for fn in os.listdir(parent):
--- a/qfieldsync/utils/file_utils.py
+++ b/qfieldsync/utils/file_utils.py
@@ -31,6 +31,7 @@ from pathlib import Path
 
 from qfieldsync.utils.exceptions import NoProjectFoundError, QFieldSyncError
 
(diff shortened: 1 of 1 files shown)


qfieldsync/utils/file_utils.py
```diff
@@ -44,6 +44,10 @@ def fileparts(fn, extension_dot=True):
 
 
 def get_children_with_extension(parent, specified_ext, count=1):
+    if not os.path.isdir(parent):
+        raise QFieldSyncError(
+            "The directory {} could not be found".format(parent))
```
m-kuhn, line 49: tr?

nirvn: We could use QCoreApplication.translate() here (as well as the other QFieldSyncErrors in this file).

nirvn: Actually maybe not, since translate can't deal with arguments. Or is this going to be a TIL moment? :)

qfieldsync/utils/file_utils.py
```diff
@@ -52,7 +57,7 @@ def get_children_with_extension(parent, specified_ext, count=1):
             res.append(os.path.join(parent, fn))
     if len(res) != count:
         raise QFieldSyncError(
-            "Expected {} children with extension {} under {}, got {}".format(
+            QCoreApplication.translate("QFieldFileUtils","Expected {} children with extension {} under {}, got {}").format(
                 count, specified_ext, parent, len(res)))
```
m-kuhn, line 61:
```suggestion
            QCoreApplication.translate("QFieldFileUtils","Expected {expected_count} children with extension {file_extension} under {parent}, got {real_count}").format(
                expected_count=count, file_extension=specified_ext, parent=parent, real_count=len(res)))
```

m-kuhn (approved):

PR #190, opened 2020-09-22
Store primary key field names as layer custom property

marioba:
Store the primary key field name(s) of offline layers as layer custom property.
The value is stored as a comma-separated list into the `QFieldSync/cloudPrimaryKeys` layer's property

### DIFF
--- a/qfieldsync/core/offline_converter.py
+++ b/qfieldsync/core/offline_converter.py
@@ -164,7 +164,7 @@ class OfflineConverter(QObject):
                     # Store the primary key field name(s) as comma separated custom property
                     if layer.type() == QgsMapLayer.VectorLayer:
                         key_fields = ','.join([layer.fields()[x].name() for x in layer.primaryKeyAttributes()])
-                        layer.setCustomProperty('QFieldSync/cloudPrimaryKeys', key_fields)
+                        layer.setCustomProperty('QFieldSync/sourceDataPrimaryKeys', key_fields)
 
                 elif layer_source.action == SyncAction.NO_ACTION:
                     copied_files = layer_source.copy(self.export_folder, copied_files)
@@ -226,7 +226,7 @@ class OfflineConverter(QObject):
                             stored_fields = original_pk_fields_by_layer_name.get(original_layer_name, None)
                             if stored_fields:
                                 layer.setCustomProperty(
-                                    'QFieldSync/cloudPrimaryKeys',
+                                    'QFieldSync/sourceDataPrimaryKeys',
                                     stored_fields)
 
                         for field in layer.fields():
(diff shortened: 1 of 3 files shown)


suricactus (approved):
Looks good to me!

marioba:
It's failing for QGIS 3.10... I'm trying to fix it

qfieldsync/core/offline_converter.py
```diff
@@ -155,6 +155,12 @@ def convert(self):
                             'Both "Area of Interest" and "only selected features" options were enabled, tha latter takes precedence.'),
                             'QFieldSync')
                     self.__offline_layers.append(layer)
+
+                    # Store the primary key field name(s) as comma separated custom property
+                    if layer.type() == QgsMapLayer.VectorLayer:
+                        key_fields = ','.join([layer.fields()[x].name() for x in layer.primaryKeyAttributes()])
```
suricactus, line 161: There is this very ugly situation when the column name could store the list separator. In this case a column with `"crazy,comma"` would fail. I think it would be better to directly give me the single column I would better use, instead of the full list. QField ignores the rest of the columns anyway:
```suggestion
                        key_field = [*layer.primaryKeyAttributes(), 'fid'][0]
                        layer.setCustomProperty('QFieldSync/cloudPrimaryKey', key_field)
```

marioba: But what happens in the case of a pk on more than one field? It's not better to check the result on the QField side for multiple values and show an error like "I cannot work with multi pks

And isn't more clear to not set the value at all (or empty string) when no pk is defined, instead of a harcoded 'fid' value, that can be wrong?

marioba: But what happens in the case of a pk on more than one field? It's not better to check the result on the QField side for multiple values and show an error like "I cannot work with multi pks"?

And isn't more clear to not set the value at all (or empty string) when no pk is defined, instead of a harcoded 'fid' value, that can be wrong?

suricactus: Multiple primary keys are a problem, true that. Currently I use the first column anyway, but I can also complain, not a big difference to me.

Assuming an attribute callled "fid" is not the best solution, but it works on memory and geojson layers, otherwise I cannot write tests.

I am ok to keep your original implementation, just add an assert that the column names do not contain the delimiter (`,` in this case).

marioba: I can also use a different delimiter than `,`...

marioba: Is there something that cannot be in the field name?

suricactus: AFAIK in postgres there is no limitation what symbols can be used as a column name

qfieldsync/core/offline_converter.py
```diff
@@ -155,6 +155,12 @@ def convert(self):
...
+                        layer.setCustomProperty('QFieldSync/cloudPrimaryKeys', key_fields)
+
```
m-kuhn, line 163: Can we rename this to "sourceDataPrimaryKey" or similar? Since it's qfieldsync (and not qfieldcloud only).

marioba: ok

PR #194, opened 2020-10-14
Move the core of the plugin into a separate library libqfieldsync

suricactus:
The motivation behind the split is in the need to use the existing QFieldSync plugin logic for other methods of export, including command line or other applicattions.

For end users the split should not affect them in any way.

For developers they should just remember to call `git submodule update`. But they already do, because `settingmanager` is also a submodule.


m-kuhn:
Very short merge window, I'd like to ask some questions here...

suricactus:
Should I revert the change? Or we can comment here or on https://github.com/opengisch/qfieldsync/tree/qfield_cloud ?

m-kuhn:
leave it like this for the moment.

1. main questions was if it makes sense to base qfieldcloud on qfieldsync
2. question was if we could use pypi instead of git submodule

suricactus:
1) I think it is much safer to build on this. I am not aware of (any) big changes needed for the cloud export, otherwise we would duplicate the code. Or at least until we stop using `QgsOfflineEditing`. QFieldCloud would use only the new `libqfieldsync` subset, which can gradually move away from what is it now.

2) The `settingmanager` was already using a submodule, so I went for this. Also it is a bit easier to start development of the plugin - just `git clone && git submodule update`, rather than installing packages. No problem objections against hacing it as published package though, especially for the cloud.

m-kuhn:
Regarding 1 I agree it's related to moving away from OE.
Regarding 2 I think it should be possible to deploy as python egg. Deploy to pypi using CI/CD has already been done for https://github.com/opengisch/pum

suricactus:
It is not clear to me, should I make a pypi module now, or do it later?

m-kuhn:
Can you talk to @3nids about the complexity of this? I think it would be very nice to have but it depends on how much time is required to set this up.

3nids:
it's pretty straightforward to push to PyPi.
Travis has dedicated deploy step. 
The effort is to produce the package, not tu push it.

PR #195, opened 2020-10-14
Split the configuration for cloud and traditional cable export

suricactus:
Peek 2020-10-14 18-04

More or less the UI I managed to produce is this. Any feedback is welcome, I am not completely happy how it looks. Notable changes:
- now there is `LayerSource.action` and `LayerSource.cloud_action`
- the image naming is moved only in the per layer settings (no more in project settings)
- the "Basemap" configuration is moved out of the tab and it is a checkable box whether to be enabled or not

A note on the "Prefer Off/Online Layers" - it will ignore toggling the layer action, if the layer is already marked as "Remove from Project", it will be annoying if it restarts the action every time.

The last remaining issue is with lock geometry, I need ideas where to move it. It is not kept in sync between the "Cloud" and "Cable" tabs, only the latter is taken. I believe we can remove it from the table and also place it in the Layer Properties dialog only. Ideas?

### DIFF
--- a/qfieldsync/gui/layers_config_widget.py
+++ b/qfieldsync/gui/layers_config_widget.py
@@ -1,0 +1,167 @@
+# -*- coding: utf-8 -*-
+"""
+/***************************************************************************
+ QFieldSyncDialog
+                                 A QGIS plugin
+ Sync your projects to QField on android
+                             -------------------
+        begin                : 2020-10-10
+        git sha              : $Format:%H$
+        copyright            : (C) 2020 by OPENGIS.ch
+        email                : info@opengis.ch
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+"""
+import os
+
+from qgis.core import QgsProject
+
+from qgis.PyQt.uic import loadUiType
+from qgis.PyQt.QtWidgets import QWidget
+
+from qgis.PyQt.QtCore import Qt, QTemporaryDir
+from qgis.PyQt.QtGui import QIcon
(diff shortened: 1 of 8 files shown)


nirvn:
Few comments:
- you'll need a vertical spacer in the cloud panel to avoid bad spacing when the advanced section is collapsed
- [x] Only Copy Feature in Area of Interest is oddly placed here
- I'd avoid using Tile Capitalization for checkbox, radio button labels

suricactus:
1) Tried that, the problem is that the radio buttons remain way too small on the top and a bit hard to find in the whole UI. 
2) I agree. Another non-collapsible groupbox ("Advanced Settings"?) around it might fix it?
3) I was following the QGIS HIG.
https://docs.qgis.org/3.10/en/docs/developers_guide/hig.html
> Capitalise first letter only in labels, tool tips, descriptive text, and other non-heading or title text: These should be written as a phrase with leading capital letter, and all remaining words written with lower case first letters, unless they are nouns
> 
> Capitalize all words in Titles (group box, tab, list view columns, and so on), Functions (menu items, buttons), and other selectable items (combobox items, listbox items, tree list items, and so on): Capitalize all words, except prepositions that are shorter than five letters (for example, ‘with’ but ‘Without’), conjunctions (for example, and, or, but), and articles (a, an, the). However, always capitalize the first and last word.

nirvn:
@suricactus ,

1. Can you post an image? ATM, this flooding in the middle of empty space is really not great
2. Is the option used both for cloud and cable?
3. Right, as far as QGIS' HIG in concerned, you'd only capitalize first letter of radiobox/checkbox labels. E.g.:
image

suricactus:
1) Is it possible to make the tab height to become as small as possible? I have played with a lot of combinations with `sizePolicy` and `layoutStretch` with no success? Or it is a UI antipattern? Here we go:
2) It is for both cloud and cable. Check the screenshot above with the additional groupbox a bit better, but still not the best.
3) Here we go:
Screenshot from 2020-10-15 11-36-33
Screenshot from 2020-10-15 11-35-41

nirvn:
It's definitively an improvement. I think we can settle on that for now.

m-kuhn:
Thanks @suricactus and @nirvn 

One note, please don't forget to cleanup after merge (especially if the branch is not done on a forked repo)

image

PR #200, opened 2020-10-26
Don't use layer suffix ` (offline)`

signedav:
Since this https://github.com/qgis/QGIS/pull/39492 the suffix ` (offline)` can be avoided to avoid issues with expressions etc.

So let's don't use it anymore.


m-kuhn (approved):

m-kuhn:
pssss, delete the branch after merge ;)

PR #204, opened 2020-12-02
[sync dialog] bootstrap color instead of named html colors

olivierdalang:
Looks slightly less 1995-ish

image


suricactus:
Thanks, @olivierdalang :)

PR #215, opened 2021-01-26
Manually trigger packaging a new .zip

suricactus:


suricactus:
Seems I need to merge to test the manual build of packages, tests pass, at least I didn't break anything

PR #216, opened 2021-01-30
Stop adding new PhotoNamingTableWidgets on reloadProject

uclaros:
Using the *toggle layers* menu in the project properties would mess with the *Photo Naming* tab, adding a new table widget every time.


suricactus:
Nice fix, thank you!

uclaros:
The logic behind re-populating this table needs some tuning. The *Toggle layers* menu entries cause changes to the *Naming expression* to be lost.

PR #218, opened 2021-01-30
Don't overexpand the photo naming table horizontally

### LINKED ISSUE #198: QField tab in layer properties is unnecesary wide
When opening the QField tab in the layer properties window, the tab selector is forced to shrink and displaying icons only, regardless of the total width of the properties window.
image

uclaros:
Fixes #198


suricactus:
Thanks @uclaros !

PR #220, opened 2021-02-24
qgis:rasterize processing tool changed it's parameters from QGIS 3.10 to QGIS 3.16

roger1233:
From QGIS 3.10 to QGIS 3.16, the qgis:rasterize processing tool changed it's parameters, specifically the "LAYER" parameter does not exist anymore, and has been replaced by the "LAYERS" parameter, expecting a list. This change fixes the QGIS 3.16 compatibility problem.

The symptom is that the raster created still exists, but instead of containing only one layer (normally the map layer) it contains by default all visible layers, therefore rasterizing not only the basemap but also the vectorial objects into it.

This change fixes the bug and makes the Qpackage compatible with QGIS 3.16 which is now the "long term stable" version.

### DIFF
--- a/qfieldsync/core/offline_converter.py
+++ b/qfieldsync/core/offline_converter.py
@@ -310,10 +310,13 @@ class OfflineConverter(QObject):
                                              self.extent.yMaximum())
 
         alg = QgsApplication.instance().processingRegistry().createAlgorithmById('qgis:rasterize')
-
+        
+        # passing the LAYER and LAYERS parameters since QGIS 3.10 uses the LAYER parameter and 3.16+ uses the LAYERS parameter instead
+        # keeping both ensures backward compatibility
         params = {
             'EXTENT': extent_string,
             'MAP_THEME': map_theme,
+            'LAYER': layer,
             'LAYERS': [layer],
             'MAP_UNITS_PER_PIXEL': map_units_per_pixel,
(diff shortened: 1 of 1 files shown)


suricactus:
Thanks for the fix! 

Is it possible to keep it still compatible with the pre-LAYERS arguments by checking for `Qgis.QGIS_VERSION_INT <= 31200`? I know QGIS 3.10.x got it's final release last month, but I guess there are some organizations that will stick to it for a while.

roger1233:
> 
> 
> Thanks for the fix!
> 
> Is it possible to keep it still compatible with the pre-LAYERS arguments by checking for `Qgis.QGIS_VERSION_INT <= 31200`? I know QGIS 3.10.x got it's final release last month, but I guess there are some organizations that will stick to it for a while.

Hey, good idea. Maybe adding the old "LAYER" parameter too would work, since it did not generate an error for me on 3.16. No need to check the version that way.

suricactus:
Even better, then just add a comment we have both parameters passed for compatibility. :)

roger1233:
Good call. Done. And tested with 3.16 working.

suricactus:
Thanks @roger1233 !

roger1233:
You're very welcome. Thank you for maintaining this wonderful tool. Happy to help, as small as it is.

PR #221, opened 2021-02-24
Add QGIS 3.16 LTR in the test matrix

suricactus:
Turns out QGIS no longer produces `final_3_XX_XX` docker tags, so I also switched to `release_3_10`.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        qgis_version: [release_3_10, release_3_16, latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous_integration.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        qgis_version: [final-3_10_9, latest]
+        qgis_version: [release_3_10, release_3_16, latest]
```
suricactus, line 26:
```suggestion
        qgis_version: [release-3_10, release-3_16, latest]
```
Whoops

m-kuhn (approved):

PR #233, opened 2021-04-20
add warning when a value relation has no layer set

3nids:
in my case, the field was not shown in the form.

a complementary fix would be to parse only shown fields


signedav (approved):
Looks good. But maybe we can only consider the relevant fields. So there are no messages for fields that are not relevant for the user anyway.

Edit: Just seen that this is what you propose on the issue. Well then. We can merge this, if time is short.

PR #246, opened 2021-05-26
WIP Additional cloud fixes

### LINKED ISSUE #240: After downloading a project the "Open project" button is not enabled
But if you refresh the project list, then it becomes enabled.
image

suricactus:
- [x] More info what is actually deleted when click on the delete project
- [x] Fix when the project was downloaded, then deleted, then downloaded again
- [x] Refresh projects list after a project is downloaded (fix #240)
- [x] Allow different owners, other than the current user
- [x] Show only your own projects, do not include public
- [x] Enforce permissions when updating a project
- [x] Fix alignment of the dynamic fields in "base layer" group box
- [x] Project extent - used in offline editing and base map export

### DIFF
--- a/qfieldsync/gui/project_configuration_widget.py
+++ b/qfieldsync/gui/project_configuration_widget.py
@@ -19,12 +19,18 @@
 """
 import os
 
-from qgis.core import QgsMapLayerProxyModel, QgsProject
-from qgis.gui import QgsOptionsPageWidget
+from qgis.core import (
+    QgsCoordinateReferenceSystem,
+    QgsMapLayerProxyModel,
+    QgsPolygon,
+    QgsProject,
+)
+from qgis.gui import QgsExtentWidget, QgsOptionsPageWidget
 from qgis.PyQt.QtCore import Qt
 from qgis.PyQt.QtGui import QIcon
 from qgis.PyQt.QtWidgets import QLabel
 from qgis.PyQt.uic import loadUiType
+from qgis.utils import iface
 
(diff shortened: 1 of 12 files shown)


qfieldsync/gui/project_configuration_widget.py
```diff
@@ -62,9 +62,11 @@ def __init__(self, parent=None):
         self.areaOfInterestExtentWidget.setToolTip(
             self.tr("Leave null to use the current project zoom extent.")
         )
-        self.areaOfInterestExtentWidget.setMapCanvas(iface.mapCanvas())
         self.areaOfInterestExtentWidget.setNullValueAllowed(True)
 
+        if iface:
```
m-kuhn, line 67: How can this be run without iface again?

suricactus: I guess when run from commandline, iface is `None`, it failed the previous job

m-kuhn: is the `gui` lib used from `cli`?

suricactus: no, none of them is using the gui:
```
import os
import shutil
import tempfile

from qgis.core import Qgis, QgsOfflineEditing, QgsProject
from qgis.testing import start_app, unittest
from qgis.testing.mocked import get_iface

from qfieldsync.libqfieldsync import OfflineConverter
from qfieldsync.tests.utilities import test_data_folder

start_app()
```

PR #252, opened 2021-06-09
Add UI to pick a digitizing logs layer

nirvn:
A new digitizing logs layer combo box was added in the project properties' QField panel:
image

This is to accompany the new feature being developed for QField 2.0, see https://github.com/opengisch/QField/pull/1971

### DIFF
--- a/qfieldsync/gui/project_configuration_widget.py
+++ b/qfieldsync/gui/project_configuration_widget.py
@@ -84,7 +84,7 @@ class ProjectConfigurationWidget(WidgetUi, QgsOptionsPageWidget):
             )
 
         self.advancedSettingsGroupBox.layout().addWidget(
-            self.areaOfInterestExtentWidget, 0, 1
+            self.areaOfInterestExtentWidget, 1, 1
         )
 
         self.preferOnlineLayersRadioButton.clicked.connect(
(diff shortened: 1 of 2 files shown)


qfieldsync/gui/project_configuration_widget.py
```diff
@@ -193,6 +202,14 @@ def apply(self):
...
         except AttributeError:
             pass
+        try:
+            self.__project_configuration.digitizing_logs_layer = (
+                self.digitizingLogsLayerComboBox.currentLayer().id()
+                if self.digitizingLogsLayerComboBox.currentLayer()
+                else ""
+            )
+        except AttributeError:
+            pass
```
suricactus, line 212: When does it throw?

suricactus (approved):

qfieldsync/gui/project_configuration_widget.py
nirvn: @suricactus , I took the same approach as the base map layer setting, assume it was there for a reason.

suricactus: I also had no idea for the above setting, hoped you found a situation when it throws. The rest of the settings are not surrounded by try/catch. Let's hope for the best. :orthodox_cross: Anyway, it is safer this way. Thanks.

Issue #256, opened 2021-06-18
python error when packaging 


3nids:
![image](https://user-images.githubusercontent.com/127259/122568305-e9490980-d049-11eb-829d-d0443546d583.png)

with 4.0 beta 6

suricactus:
Do you have QGIS 3.20? Just updated and got the same error.

3nids:
yes, master

suricactus:
Any idea why the slots might suddenly get broken?

3nids:
change of PyQt version? or you are still on the same?

suricactus:
| old | new |
|-----|-------|
| ![Screenshot from 2021-06-21 18-57-31](https://user-images.githubusercontent.com/2820439/122792726-194f1180-d2c3-11eb-8216-784690218c92.png) | ![Screenshot from 2021-06-21 18-59-36](https://user-images.githubusercontent.com/2820439/122792723-181de480-d2c3-11eb-9123-eb01f95d0da5.png) |

suricactus:
Just tested, the complete removal of the `pyqtSlot` slot made it work in 3.20.
```diff
    @pyqtSlot("QgsOfflineEditing::ProgressMode", int)
    def on_offline_editing_max_changed(self, _, mode_count):
        self.__max_task_progress = mode_count
```

PR #257, opened 2021-06-23
Kill pyqt slots for QGIS 3.20

suricactus:
Still not sure why the slots no longer work in 3.20, but they are not needed anyways.

Fixes https://github.com/opengisch/qfieldsync/issues/256


m-kuhn (approved):

3nids:
that would be worth a release no?

suricactus:
Yes, done.

PR #270, opened 2021-07-29
Improve synchornizing of cloud projects missing a local directory

nirvn:
This is the last bit of UI/UX improvement I had on my to-do list for qfieldsync's QFieldCloud support.

This IMHO greatly improves the first-time download/synchronization of a cloud project that hasn't had a local directory set up yet. The changes include:
- Instead of being thrown a directory picker dialog out of nowhere, the synchronization dialog informs the user of what's happening
- To ease the process, a default local directory is provided (derived from the cloud project name and the default cloud directory set in the user preferences)
- Instead of being shown the intimidating file tree widget on first synchronization, proceed with downloading all files directly
- At the end of a successful download/synchronization, the user is provided with a useful checkbox to automatically load the downloaded/sync'ed project after closing the dialog.

This is how the UI/UX now looks like when downloading a cloud project from the QFieldCloud server:
Peek 2021-07-29 11-59

I think with that in, we've just made opening new server-stored cloud projects much, _much_ easier.

(@suricactus , this fixes the stracktrace thrown in the browser panel)

### DIFF
--- a/qfieldsync/gui/cloud_transfer_dialog.py
+++ b/qfieldsync/gui/cloud_transfer_dialog.py
@@ -20,8 +20,10 @@
  *                                                                         *
  ***************************************************************************/
 """
+import glob
 import os
 from enum import Enum
+from pathlib import Path
 from typing import Dict, List
 
 from qgis.PyQt.QtCore import QObject, Qt, pyqtSignal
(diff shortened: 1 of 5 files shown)


m-kuhn:
Excellent

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -179,8 +224,17 @@ def prepare_project_transfer(self):
...
+
+        if self.download_project:
+            self._on_prefer_cloud_button_clicked()
+            self.on_project_apply_clicked()
+        else:
+            self.build_files_tree()
+            self.stackedWidget.setCurrentWidget(self.filesPage)
+            self.buttonBox.button(QDialogButtonBox.Apply).setVisible(True)
+            self.buttonBox.button(QDialogButtonBox.Apply).setText(
+                self.tr("Download project")
```
suricactus, line 236: I am now in a weird situation I just created a new project and it says "Download project"

nirvn: Something must not register properly when you created the project re local_dir? That'd be the only way download appears here.

suricactus: Why not "synchronize project" as neutral?

suricactus: I created a project and set the local dir during creation.

nirvn: I felt it was better to communicate to the user the difference between synching pre-existing local <-> remote vs. populating local (i.e. downloading).

PR #279, opened 2021-08-02
Warn if the layer contains memory layers

### LINKED ISSUE #222: When a project contains temprorary scratch layers, warn the user
Temporary scratch layers are gone once you close QGIS. Some users are used to third-party plugins that automatically save the memory layers which worsens the situation. With the upcomming cloud changes this becomes even bigger concern, since the export is not happening in the current QGIS session.

Some popup like this might help:
```
QFieldSync has detected termporary  layers in you project which cannot be availble on other devies. Watcha gonna do?

Remove from the project
Store in a geopackage
Save the project anyway
```

suricactus:
Resolves #222 .

image

### DIFF
--- a/qfieldsync/gui/cloud_transfer_dialog.py
+++ b/qfieldsync/gui/cloud_transfer_dialog.py
@@ -26,6 +26,7 @@ from enum import Enum
 from pathlib import Path
 from typing import Dict, List
 
+from qgis.core import QgsProject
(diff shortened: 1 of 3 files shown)


m-kuhn (approved):

nirvn:
I was 1sec away from requesting for the label to be moved below the progress bar. It's position atm is overly aggressive.

nirvn:
(Ping @suricactus on that last comment)

PR #286, opened 2021-08-05
Clodify converter improvements

suricactus:
UI/UX:
- suggest a unique project name
- suggest a unique local project directory
- add the opened project in "recent projects" history
- keep a backup of the original project
- append ` (QFieldCloud)` in the converted project
- set the title on the original project, if missing

Code:
- always prefer pathlib
- fix a few static analysis issues
- `folder` -> `directory`, sorry for the strong opinion on this :)
- python naming convention

### DIFF
--- a/qfieldsync/core/cloud_converter.py
+++ b/qfieldsync/core/cloud_converter.py
@@ -38,7 +38,7 @@ class CloudConverter(QObject):
     def __init__(
         self,
         project: QgsProject,
-        export_folder: str,
+        export_dirname: str,
     ):
 
         super(CloudConverter, self).__init__(parent=None)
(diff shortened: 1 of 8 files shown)


qfieldsync/core/cloud_converter.py
```diff
@@ -135,10 +136,10 @@ def convert(self) -> None:  # noqa: C901
 
             # TODO whatcha gonna do if QgsProject::read()/write() fails
             if is_converted:
-                self.project.read(project_path)
+                iface.addProject(project_path)
```
nirvn: @suricactus , I was wondering whether we want to avoid using iface in /core. No strong opinion. 

As we discussed yesterday, +1 to rely on addProject along the way in this process.

nirvn:
@suricactus , only one comment. Looking good. All we talked about being addressed is sweet :)

qfieldsync/core/cloud_converter.py
suricactus: Hey, @nirvn . I also feel a bit nervous depending on `iface`, but I don't think there are a lot of other options.

PR #288, opened 2021-08-05
QFieldCloud integration minor fixes

suricactus:
Minor, but timeconsuming...

Check the commit messages to see what have changed.

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -612,20 +612,37 @@ class CloudProjectsCache(QObject):
 
     @property
     def is_currently_open_project_cloud_local(self) -> bool:
-        project_dir = QgsProject.instance().homePath()
+        """Checks whether the currently opened QGIS project is a configured cloud project.
 
-        found = False
+        NOTE there is a difference with `currently_opened_project()`, as this method does not
+        depend on downloaded project list.
+
+        Returns:
+            bool: opened QGIS project is configured cloud project
+        """
+        project_dir = QgsProject.instance().homePath()
+        project_ids = [p.id for p in self.projects] if self.projects else []
+
         for project_id, local_dir in self.preferences.value(
             "qfieldCloudProjectLocalDirs"
         ).items():
-            if local_dir and Path(local_dir) == Path(project_dir):
-                found = True
-                break
+            if project_ids and project_id not in project_ids:
+                continue
 
-        return found
+            if local_dir and Path(local_dir) == Path(project_dir):
+                return True
+
+        return False
 
     @property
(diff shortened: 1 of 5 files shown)


qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -804,7 +803,7 @@ def select_local_dir(self) -> Optional[str]:
         initial_path = (
             self.localDirLineEdit.text()
             or str(Path(QgsProject.instance().homePath()).parent)
-            or self.default_local_dir
+            or self.preferences.value("cloudDirectory")
```
nirvn, line 806: Excellent.

qfieldsync/core/cloud_api.py
```diff
@@ -610,20 +610,6 @@ def projects(self) -> Optional[List[CloudProject]]:
     def error_reason(self) -> str:
         return self._error_reason
 
-    @property
-    def is_currently_open_project_cloud_local(self) -> bool:
```
nirvn, line 614: @suricactus , it is _not_ the same as currently_open_project. The latter *requires* a connected state and a valid project cache, whereas this function only checks whether the local directory is registered as a mirror of a server cloud project.

Without this, the qfieldsync toolbar's cloud synchronization button can't be enabled / disabled properly when not connected, which isn't great UX.

qfieldsync/qfield_sync.py
```diff
@@ -145,6 +145,11 @@ def __init__(self, iface):
             self.cloud_item_gui_provider
         )
 
+        # autologin
+        dialog = CloudLoginDialog(self.network_manager)
```
nirvn, line 150: -1 to auto login. It's intrusive, and potentially a security risk in some environment. 

If we think an auto login is useful for some, let's lock that behavior behind a [x] auto login on QGIS launch setting.

qfieldsync/core/cloud_api.py
suricactus: Ah, I see now. You are right. A comment can help next time to highlight the difference.

suricactus: What I see as a problem is there might be a `project_id` configuration in the local paths that no longer exists.

suricactus: Added docs and a check for the project_id, if projects are present

qfieldsync/qfield_sync.py
suricactus: I will open another PR with this.

suricactus: see https://github.com/opengisch/qfieldsync/pull/290

PR #290, opened 2021-08-06
Add option to auto login

### LINKED ISSUE #284: QFieldCloud login dialog show in the background and it is easy to miss
When clicking on the "login" button, seems there is nothing happening. But there is the login dialog poped-up in the background.

suricactus:
After a quick discussion with @m-kuhn seems the "remember me" checkbox was exactly that, it was not doing the login. You will only be logged in on next session if you logged in and you have the possibility to log out. If someone intentionally never logs in, then all is good.

Enabled by default. 

Added a tooltip and changed it to "stay logged in":
image

Also fix #284

### DIFF
--- a/qfieldsync/qfield_sync.py
+++ b/qfieldsync/qfield_sync.py
@@ -145,6 +145,12 @@ class QFieldSync(object):
             self.cloud_item_gui_provider
         )
 
+        # autologin
+        if self.preferences.value("qfieldCloudRememberMe"):
+            dialog = CloudLoginDialog(self.network_manager)
+            dialog.authenticate()
+            dialog.hide()
+
     # noinspection PyMethodMayBeStatic
     def tr(self, message):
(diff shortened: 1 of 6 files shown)


qfieldsync/qfield_sync.py
```diff
@@ -145,6 +145,12 @@ def __init__(self, iface):
             self.cloud_item_gui_provider
         )
 
+        # autologin
+        if self.preferences.value("qfieldCloudRememberMe"):
+            dialog = CloudLoginDialog(self.network_manager)
+            dialog.authenticate()
```
m-kuhn, line 151: Would be nice to refactor this logic out of the dialog.

m-kuhn (approved):

qfieldsync/qfield_sync.py
suricactus: Agree

PR #293, opened 2021-08-07
Tweaks to the cloud synchronization and projects dialogs

nirvn:
@suricactus , as discussed, this PR moves the memory layer warning to sit just above the standard buttons box.

The commit also hides the error label by default until there's an actual string in there. That removes the visually awkward spacing between the collapsing details box and the download progress bar.

Finally, the PR also improves the cloud projects dialog. Here's how it looks now:
image

Improvements include:
- Add a convert current project to cloud project button
- Rework margins to save some white space and harmonize with other dialogs across QGIS
- Move the create blank project away from the projects list header into the newly introduced new cloud projects actions section and transform into into a push button so text appears on the button

@mbernasocchi , as suggested during the lightning talk, the convert current project action was removed from the toolbar in favor of being located in the cloud projects dialog. It remains available in the qfieldsync plugin sub-menu.

### DIFF
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -26,7 +26,7 @@ from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 from PyQt5.QtCore import QUrl
-from qgis.core import QgsProject
+from qgis.core import Qgis, QgsProject
 from qgis.PyQt.QtCore import (
     QDateTime,
     QItemSelectionModel,
@@ -47,6 +47,7 @@ from qgis.PyQt.QtWidgets import (
     QAction,
     QCheckBox,
     QDialog,
+    QDialogButtonBox,
     QFileDialog,
     QHBoxLayout,
     QHeaderView,
@@ -65,6 +66,7 @@ from qgis.utils import iface
 from qfieldsync.core import Preferences
 from qfieldsync.core.cloud_api import CloudException, CloudNetworkAccessManager
 from qfieldsync.core.cloud_project import CloudProject, ProjectFile, ProjectFileCheckout
+from qfieldsync.gui.cloud_converter_dialog import CloudConverterDialog
 from qfieldsync.gui.cloud_login_dialog import CloudLoginDialog
 from qfieldsync.gui.cloud_transfer_dialog import CloudTransferDialog
 from qfieldsync.utils.cloud_utils import closure, to_cloud_title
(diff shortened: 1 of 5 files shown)


qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -123,18 +125,20 @@ def __init__(
                 parent=self,
             )
         else:
-            self.welcomeLabelValue.setText(
-                self.network_manager.auth().config("username")
+            self.welcomeLabel.setText(
+                self.tr("Greetings {}.").format(
```
suricactus, line 129: I would love to keep the username in bold.

nirvn: I was aiming at keeping bold for subsections. If you're passionate about the bold here I can re-add. I'm -0, if you're +1 by my count you win ;)

nirvn: I would love to see the avatar on the top right, something to add in our respective "if we ever get free time" list ;)

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -226,9 +241,14 @@ def current_cloud_project(self) -> Optional[CloudProject]:
...
         self.update_project_table_selection()
+        self.update_ui_state()
 
     def on_auth_accepted(self):
-        self.welcomeLabelValue.setText(self.network_manager.auth().config("username"))
+        self.welcomeLabel.setText(
+            self.tr("Greetings {}.").format(
+                self.network_manager.auth().config("username")
+            )
+        )
```
suricactus, line 251: Can we add a function "update_welcome_label", which has the following logic:
```
if logged_in:
  self.welcomeLabel.setText("Greeting <b>{}</b>").format(self.network_manager.auth().config("username"))
  self.welcomeLabel.setToolTip("...").format(self.network_manager.url)
else:
  self.welcomeLabel.setText("Not logged in")
  self.welcomeLabel.setToolTip("")
```

qfieldsync/gui/cloud_projects_dialog.py
suricactus: keep it like this then, let's see how my free saturday time goes :)

suricactus (approved):
Awesome work, feel free to merge when ready.

nirvn:
@suricactus , thanks for the review. I'll merge now and apply the update welcome label suggestion in a follow up PR later tonight or tomorrow.

PR #295, opened 2021-08-07
More love to the cloud projects dialog

nirvn:
Feeling inspired:
image

Improvements include:
- massive de-cluttering of the table widget by moving the repeated actions tool button out of the table
- the set of actions now sit under the table widget (much like QGIS does in many other places) -> harmony
- relying on QGIS icons to both fit better in the overall QGIS desktop environment (bonus: we decrease icon maintenance burden on our side)


nirvn:
@suricactus , your update_welcome_label suggestion is implemented.

suricactus (approved):
:rocket:

PR #297, opened 2021-08-08
Add community projects to the cloud projects dialogs

nirvn:
GIF time:
Peek 2021-08-08 11-23

Improvements include:
- the cloud projects dialog now allow users to see and download / synchronize community projects (making the QGIS desktop experience closer to that of QField)
- relocating the refresh button to avoid awkward look following @suricactus 's fantastic avatar work
- renamed the browser panel QFieldCloud's "public projects" to "community" so we share the same keyword across QField, QGIS, and the web app

### DIFF
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -132,7 +132,6 @@ class CloudProjectsDialog(QDialog, CloudProjectsDialogUi):
         )
 
         self.projectsType.addItem(self.tr("My projects"))
-        self.projectsType.addItem(self.tr("Community"))
         self.projectsType.setCurrentIndex(0)
         self.projectsType.currentIndexChanged.connect(lambda: self.show_projects())
 
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -131,6 +131,10 @@ class CloudProjectsDialog(QDialog, CloudProjectsDialogUi):
             )
         )
 
+        self.projectsType.addItem(self.tr("My projects"))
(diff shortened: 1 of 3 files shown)


suricactus (approved):
Awesome! However, I would add one critical warning before we merge. Imagine you are organization X and upload a project that has connection to an online PostGIS layer. You say this project is public and all the users are the role "reader" (read-only) by default. However, since the layer is online, everyone who downloads the original project will have access to your DB. 

We are missing a big, phat warning saying -> hey, you are about to put your online layer accessible to the public, are you really, really sure you want this?

I know this issues exists for a while, but this PR is a good opportunity to fix it.

nirvn:
@suricactus , I _do_ think we should see read-only / non-modifiable projects in there. There's nothing wrong with having access to read-only community projects. It's a good way to show case all sorts cartographic styles, advanced form setups, etc.

I don't think the UI/UX fix lays on this end of the story.

If I toggle this [x] private checkbox, will it make my project(s) visible in the community tab of QField, the browser panel community node, and now the projects dialog?
image

If so, this is where we need a confirmation box (irrespective of whether the project will leak online layers that aren't set up to be offlined by the QFieldCloud server or not).

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -131,6 +131,11 @@ def __init__(
             )
         )
 
+        self.projectsType.addItem(self.tr("My projects"))
+        self.projectsType.addItem(self.tr("Community"))
```
nirvn, line 135:
```suggestion
```

nirvn:
@suricactus , community item removed from combo box for now. We definitively need a powwow on this aspect of QFieldCloud ASAP :)

PR #306, opened 2021-09-06
Improve the creating a new project experience with a wizard and explanations

suricactus:
A few improvements in this PR:
- There is a new, simpler "Create project" button with wizardish look and feel. An explanation what is going to happen is also present to the users.
- The create empty cloud button now has a dedicated UI
Peek 2021-09-06 17-56
- A simplified "Edit project" form, setting the public/private flag or the owner is now only a web UI thing.
- A new "Edit on QFieldCloud" button that opens up a browser
image
- When selecting a new project from the list, the table ensures the line is visible


qfieldsync/ui/cloud_create_project_widget.ui
```diff
@@ -0,0 +1,381 @@
...
+             </font>
+            </property>
+            <property name="text">
+             <string>A new QFieldCloud-compatible project is created from the currently opened QGIS project. In order to do so, datasets will be copied into an export directory that will act as your local mirror. Vector datasets will be converted to geopackage format to facilitate data synchronization from multiple devices. Other files used by the project will be copied in the new project location. You will automatically be switched to the newly created cloud project.</string>
+            </property>
+            <property name="wordWrap">
+             <bool>true</bool>
+            </property>
+            <property name="textInteractionFlags">
+             <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
```
nirvn, line 72: I don't think we need/want this text to be selectable no?

qfieldsync/ui/cloud_create_project_widget.ui
```diff
@@ -0,0 +1,381 @@
...
+          </item>
+          <item>
+           <widget class="QLabel" name="cloudifyInfoLabel">
+            <property name="font">
+             <font>
+              <italic>true</italic>
+             </font>
+            </property>
+            <property name="text">
+             <string>A new QFieldCloud-compatible project is created from the currently opened QGIS project. In order to do so, datasets will be copied into an export directory that will act as your local mirror. Vector datasets will be converted to geopackage format to facilitate data synchronization from multiple devices. Other files used by the project will be copied in the new project location. You will automatically be switched to the newly created cloud project.</string>
```
nirvn, line 66:
```suggestion
             <string>A new QFieldCloud-compatible project is created from the currently opened QGIS project. In order to do so, datasets will be copied into an export directory that will act as your local mirror. Vector datasets will be converted to geopackage format to facilitate data synchronization from multiple devices while other dataset types will be copied in the new project location.</string>
```

qfieldsync/ui/cloud_create_project_widget.ui
```diff
@@ -0,0 +1,381 @@
...
+             </font>
+            </property>
+            <property name="text">
+             <string>A new QFieldCloud project will be created, similarly to the web interface. You will be responsible to move all the project-related files in the project root directory, in order to make the files available on QFieldCloud. No project files will be uploaded until you click the sync button. Make sure the selected directory is empty or contains no more than one QGIS project file.</string>
+            </property>
+            <property name="wordWrap">
+             <bool>true</bool>
+            </property>
+            <property name="textInteractionFlags">
+             <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
```
nirvn, line 103: Same as above, I think we can get rid of those interactive flags.

qfieldsync/ui/cloud_create_project_widget.ui
```diff
@@ -0,0 +1,381 @@
...
+          </item>
+          <item>
+           <widget class="QLabel" name="createCloudInfoLabel">
+            <property name="font">
+             <font>
+              <italic>true</italic>
+             </font>
+            </property>
+            <property name="text">
+             <string>A new QFieldCloud project will be created, similarly to the web interface. You will be responsible to move all the project-related files in the project root directory, in order to make the files available on QFieldCloud. No project files will be uploaded until you click the sync button. Make sure the selected directory is empty or contains no more than one QGIS project file.</string>
```
nirvn, line 97:
```suggestion
             <string>A new blank QFieldCloud project instance will be created. You will be responsible to move all the project-related files within a local directory, with the project file at its root. No project files will be uploaded until you click the synchronize button. Make sure the selected directory contains no more than one QGIS project file.</string>
```

Creating a blank QFieldCloud project doesn't create a blank .qgs in the local directory, tried to rephrase things a bit so people understand they will also have to copy/create a project file in there too.

qfieldsync/ui/cloud_create_project_widget.ui
m-kuhn: copied *to*?

qfieldsync/ui/cloud_create_project_widget.ui
m-kuhn: Does this "new project instance" already define the root folder? If yes, the wording should reflect that (instead of "*a* local directory" which is very unspecific)
Then I'd probably revert the wording to positive "Project files will be uploaded when you click the synchronize button".

nirvn: AFAIK, it doesn't. It merely creates a blank server-side cloud project. The project preferences panel subsequently allows you to pick a local folder.

suricactus: @nirvn , no, you can select the directory as you create the blank project on the cloud. So the workflow is like this:
1) select "create a new empty cloud project", click "Next"
2) choose a project name **and** project local directory. Press "Submit" and if all wents well on the server, that cloud project is associated with that local dir.
3) then you need to press "Synchronize", which will ask whether to upload all the files within the local directory to the cloud.

So a final proposal:

> A new blank QFieldCloud project will be created. You will be responsible to move all the project-related files within the selected local directory, with the project file at its root. Project files will only be uploaded when you click the synchronize button. Make sure the selected directory contains no more than one QGIS project file."

qfieldsync/ui/cloud_create_project_widget.ui
suricactus: Both suggestions accepted.

qfieldsync/ui/cloud_create_project_widget.ui
nirvn: @suricactus , did you make the local directory mandatory pre pressing "submit"?

qfieldsync/ui/cloud_create_project_widget.ui
suricactus: Agreed.

qfieldsync/ui/cloud_create_project_widget.ui
suricactus: it is not mandatory, but it is prefilled with the currently open project path already

nirvn: OK -- are we doing a check against currently open project path to make sure it's a valid one? I.e., say someone wants to create a blank cloud project while he's in a QGIS session with another non-cloud project open from a directory that contains multiple project files, etc.

suricactus: I was planning to add this in a follow up PR. The code already exists and checks it when you are modifying the project, but not whn you are creating a new one. I will add it now.

suricactus: Okay, now the feedback is there 
image

nirvn (approved):

suricactus:
Wait for merging it myself.

PR #321, opened 2021-09-20
Fixes while updating the project details

suricactus:
- Disable "Is Private" checkbox in the project edit form. It is only editable using the web UI
- Keep the "Submit" button enabled even if the local project path is empty
- Do not update the local project path while editing, wait for "save" to be clicked
- Skip API update if there is nothing to be updated (name, descr). Allow updating project local path if you don't have permission to do so.
- Add .qt_for_python to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,3 +168,4 @@ qfield-sync.*.zip
 qfieldsync/resources_rc.py
 
 .vscode/**
+.qt_for_python/**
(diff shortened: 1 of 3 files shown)


nirvn (approved):

mbernasocchi:
Task linked: [DEV-3144 Fix updating project details when you are just a reader](https://app.clickup.com/t/2192114/DEV-3144)

PR #325, opened 2021-09-29
Update libqfieldsync to catch an important fix for QGIS 3.16 windows users

nirvn:


nirvn:
(more information on this here https://github.com/opengisch/libqfieldsync/pull/19 )

PR #329, opened 2021-09-30
Fix the icons without changing the current workdir globally

suricactus:
The old solution cause the whole python environment in QGIS to change.
This will break so many scripts!  Fixes the fix in #327

### DIFF
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -20,6 +20,7 @@
  *                                                                         *
  ***************************************************************************/
 """
+import os
 from pathlib import Path
 from typing import Callable, Optional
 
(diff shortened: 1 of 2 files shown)


m-kuhn (approved):
Looks good

m-kuhn:
Use this construct for extra karma points

```
class CustomCwd(object):
    def __init__(self, cwd):
        self.old_cwd = os.getcwd()
        os.chdir(cwd)
  
    def __exit__(self):
        os.chdir(self.old_cwd)


with CustomCwd(str(Path(....))):
    self.loadUi()
```

suricactus:
Or even better - a decorator :)

m-kuhn:
Decorate the constructor? Or something else? Whatever works for you :-)

PR #331, opened 2021-10-01
Cloud synchronization improvements

nirvn:
This PR improves the cloud synchronization process through two changes.

First, if users are synchronizing a project where they are assigned the role of 'reader', the UI will disable all upload-related functionality, simplifying the UI quite a lot:
image

The second is IMHO an important UX improvement: instead of always forcing uploads by default, the synchronization dialog will pick a better default based on last modified date. If the cloud-stored file has a modification date that's newer than the locally stored file, we now default to downloading from the cloud.

### DIFF
--- a/qfieldsync/core/cloud_project.py
+++ b/qfieldsync/core/cloud_project.py
@@ -62,14 +62,14 @@ class ProjectFile:
         if not self.versions:
             return
 
-        return self.versions[-1].get("created_at")
+        return self.versions[-1].get("last_modified")
 
     @property
     def updated_at(self) -> Optional[str]:
         if not self.versions:
             return
 
-        return self.versions[-1].get("updated_at")
+        return self.versions[0].get("last_modified")
 
     @property
     def versions(self) -> Optional[List[Dict[str, str]]]:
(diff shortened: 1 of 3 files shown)


mbernasocchi:
Task linked: [DEV-3282 If READER collaborator tries to upload a project file, he knows he can way too late when he uploads a file. Disable the "use local" checkbox.](https://app.clickup.com/t/2192114/DEV-3282)

mbernasocchi:
Task linked: [DEV-3296 Do not ask to reopen the project after sync if there was no downloading action.](https://app.clickup.com/t/2192114/DEV-3296)

mbernasocchi:
Task linked: [DEV-3294 Rename "prefer none" to "deselect all actions"](https://app.clickup.com/t/2192114/DEV-3294)

qfieldsync/core/cloud_project.py
```diff
@@ -62,14 +62,14 @@ def created_at(self) -> Optional[str]:
         if not self.versions:
             return
 
-        return self.versions[-1].get("created_at")
+        return self.versions[-1].get("last_modified")
```
suricactus, line 65: Ah, good I already had this in the API

suricactus: But why for "created_at"?

suricactus (approved):

PR #332, opened 2021-10-02
Do not pretend there is a project on completely empty cloud project

nirvn:
Title says it all. With this PR, the user will be aware of completely empty cloud project:

image


mbernasocchi:
Task linked: [DEV-3330 When trying to download an empty project from the cloud, it is not clear that it is empty. The synnc button is still active and when the downlaod finishes, it proposes to open the project.](https://app.clickup.com/t/2192114/DEV-3330)

suricactus (approved):

PR #333, opened 2021-10-03
Add missing cloud project icon in the browser panel and projects overview dialog

nirvn:
The browser panel's QFieldCloud node looked half broken without an icon for project items, this PR remedies to that. 

This also allows us to further harmonize things across QField and qfieldsync by making use of the remote(-only) cloud project icon we already have in QField within the browser panel and the projects overview to show whether a cloud project is available locally or not.

This is how the browser panel looks like with the PR:
image

And the projects overview dialog (note how the whole local column + checkboxes are gone in favor of the icon state):
image


nirvn:
Here's a side-by-side qfieldsync <-> QField shot:
image

Always good to increase the shared visual language between QField and the QGIS qfieldsync plugin.

qfieldsync/gui/cloud_browser_tree.py
```diff
@@ -180,6 +180,18 @@ def __init__(self, parent, project):
             "/QFieldCloud/project/" + project.id,
         )
         self.project_id = project.id
+        project = parent.network_manager.projects_cache.find_project(self.project_id)
+        self.setIcon(
```
suricactus, line 184: This might need https://github.com/opengisch/qfieldsync/pull/329 on windows 3.16.

suricactus, line 184: This might need https://github.com/opengisch/qfieldsync/pull/329 on windows 3.16.

qfieldsync/ui/cloud_projects_dialog.ui
```diff
@@ -155,14 +155,6 @@
...
           </property>
          </column>
-         <column>
-          <property name="text">
-           <string>Local</string>
-          </property>
-          <property name="toolTip">
-           <string>Whether the file has a configured local directory</string>
-          </property>
-         </column>
```
suricactus, line 165: I would keep the tooltip that explains what the icon means somehow.

nirvn: You mean adding a tooltip for the name column? (I've kept a tooltip when you mouse hover the row)

suricactus (approved):
Nice, nice, NICE!

A few thinks we can do in addtion:
- keep the tooltip that explains what is the meaning of the icon
- check if the icon thing works on QGIS 3.16 on windows (can be follow-up PR)
- in QField we list all the projects that are in the `cloud_projects` directory, including those that have no Cloud equvalent. Just a reminder, I don't have any opinion whether we should do it like this on QFS.

qfieldsync/ui/cloud_projects_dialog.ui
suricactus: Shouldn't it possible to have a tooltip on the icon itself?

nirvn: AFAIK, no as it's merely a QIcon added as a Qt.DecorationRole in the item. But again, we now have the full name column for each row with the tooltip, hit area is much larger.

nirvn:
Merging, let's follow up in additional PRs if needed. We need a new beta! :-)

PR #337, opened 2021-10-08
Bump QFieldSync to fix too many warnings when non UTF8 and non ASCII is used

### LINKED ISSUE #322: Message to rename ACSII encode filename
When using QGis 3.16-3 Hannover with the latest experimental QFieldSync version v4.0.0-beta13 for cloud sync, I got the following error message (for all the layers in the project) when making a simple package for copying manually to another device.
When installing an older version of the QField plug-in (v3.4.4), there is no such error.

PrintscreenProbleem

Could this problem also cause https://github.com/opengisch/qfieldcloud/discussions/81 ?

### LINKED ISSUE #335: warnings about layers not using UTF-8
with last beta release, I am getting a dozen of warnings about layers not using UTF-8

image

I haven't looked much into this yet, but having to click so many times is quite a bad experience

suricactus:
Fix #335
Fix #322 

Depends on https://github.com/opengisch/libqfieldsync/pull/20


duke-nyuki:
Task linked: [DEV-3387 Fix too many alerts when the layers are not in ASCII or UTF-8](https://app.clickup.com/t/2192114/DEV-3387)

PR #339, opened 2021-10-15
Bump libqfieldsync with fixes on basemap creation; Fix assert on sync

suricactus:


duke-nyuki:
Task linked: [CU-1mrf4wu Fix basemap creation throwing an error if no basemap layer is selected](https://app.clickup.com/t/1mrf4wu)

PR #342, opened 2021-11-16
UX when using the transferring dialog

suricactus:
- Use CloudTransferDialog.show_transfer_dialog static method to prevent multiple dialogs
- If the updating the files list is unsuccessful before transferring, show an error message, instead of silently showing all the local files as changed.

image


duke-nyuki:
Task linked: [QF-1413 Debug downloading `files` list is empty on GET `/files`](https://app.clickup.com/t/2192114/QF-1413)

PR #348, opened 2021-12-12
Fix a python error when transfering files

suricactus:
Hard to reproduce, but debugger pointed to here and now I am sure it is fixed.


duke-nyuki:
Task linked: [QF-1509 Transfering files in QFieldSync results in an error](https://app.clickup.com/t/2192114/QF-1509)

nirvn (approved):

PR #352, opened 2022-01-19
Don't show successfully synced message if there were some error during sync

suricactus:
After:
image


duke-nyuki:
Task linked: [QF-1630 Don't show successfully synced message if there were some error during sync](https://app.clickup.com/t/2192114/QF-1630)

PR #363, opened 2022-01-21
Actually set dirname, not only the feedback

suricactus:


duke-nyuki:
Task linked: [QF-1639 Local path is not set to current project when creating a new project](https://app.clickup.com/t/2192114/QF-1639)

PR #364, opened 2022-01-21
Fix inverted logic and weird feedback when a project is cloudified

### LINKED ISSUE #349: Cannot upload local QfieldSync directory
> I have a bug when trying to convert a project into cloud project
> 
> When I want to convert currently open project to cloud project, if I use the button with three dots, then select one folder in my computer, I have a message saying that "The entered path does not contain a QGIS project file yet" then I click "ok" and it send me back to the folder selection (see image below)
> 
> cannot upload local QfieldSync directory
> 
> However, if I write the folder path by hand instead of using the three dots button, then the button "create" become enabled. 
> 
> manual entry of the directory

Initially reported as [#185](https://github.com/opengisch/qfieldcloud/issues/185) by @MMR100

suricactus:
Fix #349

### DIFF
--- a/qfieldsync/utils/cloud_utils.py
+++ b/qfieldsync/utils/cloud_utils.py
@@ -81,7 +81,7 @@ def local_dir_feedback(
         message = dummy.tr("The entered path does not contain a QGIS project file yet.")
         status = no_project_status
 
-        if single_project_status is not LocalDirFeedback.Success:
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-1514 Local directory path feedback is weird when "cloudify" project is selected](https://app.clickup.com/t/2192114/QF-1514)

PR #365, opened 2022-01-22
Restore delete functionality

suricactus:
Until a better solution is found, better restore what was already working well enough.


duke-nyuki:
Task linked: [QF-1506 There is no way to delete files on the cloud once uploaded](https://app.clickup.com/t/2192114/QF-1506)

PR #368, opened 2022-01-23
Fix deleting remote files

suricactus:


duke-nyuki:
Task linked: [QF-1632 QFieldSync transfer log does not show any feedback when deleting local files](https://app.clickup.com/t/2192114/QF-1632)

PR #369, opened 2022-01-23
Keep local dirs configuration even if you change the API url

suricactus:


duke-nyuki:
Task linked: [QF-1642 Local dir configuration lost after QFC URL is changed](https://app.clickup.com/t/2192114/QF-1642)

PR #370, opened 2022-01-23
Disable the default timeout of 60 secs that QGIS has

suricactus:
Fix https://github.com/opengisch/qfieldcloud/issues/68


duke-nyuki:
Task linked: [QF-1649 Timeout errors in QFieldSync when uploading large files](https://app.clickup.com/t/2192114/QF-1649)

PR #373, opened 2022-01-26
Fix sync dialog when dowloading project without selected local dir

suricactus:


duke-nyuki:
Task linked: [QF-1666 Cloudify not-implemented](https://app.clickup.com/t/2192114/QF-1666)

PR #374, opened 2022-01-26
Fix logout button missing for avatar-less accounts

nirvn:
This PR fixes the logout button (now merged with the avatar display) going missing for avatar-less cloud accounts.

Our beautiful favorite bee acts as the avatar when missing:
image


suricactus:
agreed

PR #375, opened 2022-01-26
Always use the singleton QgsNetworkAccessManager instance, otherwise ssl handling is impossible from python

suricactus:
@nirvn should fix your issue with self-signed or unknown authority certificate.


duke-nyuki:
Task linked: [QF-1668 Fix connection to server with invalid SSL](https://app.clickup.com/t/2192114/QF-1668)

nirvn (approved):
That fixes connection issues with the custom qfieldcloud instance, perfect.

PR #377, opened 2022-01-27
Worst typo in history breaks downloading

suricactus:
Because autocomplete.


nirvn:
Hm funny, fixed that and never pushed the commit into a PR. Duh!

PR #378, opened 2022-01-27
Don't block the main thread when cloudifying a project so the UI updates

suricactus:
@nirvn can you confirm it works fine? I tested with two (small) projects and seems to be much better UX.

### DIFF
--- a/qfieldsync/gui/cloud_create_project_widget.py
+++ b/qfieldsync/gui/cloud_create_project_widget.py
@@ -27,7 +27,7 @@ from typing import Optional
 
 from qgis.core import Qgis, QgsProject
 from qgis.gui import QgisInterface
-from qgis.PyQt.QtCore import QDir, QRegularExpression, Qt, pyqtSignal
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-1672 Don't block the main thread when cloudifying a project](https://app.clickup.com/t/2192114/QF-1672)

qfieldsync/gui/cloud_create_project_widget.py
```diff
@@ -168,16 +171,19 @@ def cloudify_project(self):
...
+                critical_message = self.tr(
+                    "The project could not be converted into the export directory."
+                )
+                self.iface.messageBar().pushMessage(critical_message, Qgis.Critical, 0)
+                self.close()
+                return
+
+        self.timer.singleShot(0, convert)
 
         self.create_cloud_project()
```
m-kuhn, line 188: Is it a problem that line 188 is now done before create is executed?

suricactus: If you have a fast internet - no :) Thanks for spotting it.

nirvn (approved):
Logic is now sound ;)

PR #379, opened 2022-01-27
When cloudifying a project, auto set the approprite cloud action to avoid synchronization dialog errors

nirvn:
The project cloudifier should be as friction-less as possible; set good default cloud actions to converted layers so when users hit the synchronize button, they aren't shown a scary dialog full of errors :)

Prevents this dialog from showing:
image

@jmad1v07 , this is something you were hitting while doing testing.

### DIFF
--- a/qfieldsync/core/cloud_converter.py
+++ b/qfieldsync/core/cloud_converter.py
@@ -106,8 +106,11 @@ class CloudConverter(QObject):
                             ).format(layer.name()),
                         )
                         self.project.removeMapLayer(layer)
+                    else:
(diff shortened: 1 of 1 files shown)


suricactus:
Yes, I do agree. Thanks!

suricactus:
Well, I wanted to check one more thing... Why don't you use `LayerSource.default_cloud_action`?

nirvn:
@suricactus , I was looking for this constant in the wrong place and couldn't find. I'll adjust tomorrow.

PR #383, opened 2022-02-07
Fix QFieldSync package dialog if no project is open

suricactus:
It was hanging forever :(


duke-nyuki:
Task linked: [QF-1716 Fix QFieldSync package dialog if no project is open](https://app.clickup.com/t/2192114/QF-1716)

PR #386, opened 2022-02-08
QFieldSync should use the version display format from as provided from the API

suricactus:
The comment and the default value should be removed in the future.

image


duke-nyuki:
Task linked: [QF-1729 QFieldSync should use the version display format from as provided from the API](https://app.clickup.com/t/2192114/QF-1729)

PR #390, opened 2022-02-22
Fix failing login due to multiple /api/v1/ path appending

suricactus:


duke-nyuki:
Task linked: [QF-1764 Login to QFieldCloud from plugin is not working](https://app.clickup.com/t/2192114/QF-1764)

PR #395, opened 2022-03-05
If the request is not successful, do not download the response contents

suricactus:
Also the early return was causing the UI to stay in an undefined state.


duke-nyuki:
Task linked: [QF-1800 QFIeldSync should not store the downloaded file on error](https://app.clickup.com/t/2192114/QF-1800)

PR #401, opened 2022-03-06
Make sure the suggested project name is valid when creating a new project on QFIeldCloud

suricactus:
We didn't escape if the project name is generated from the filename.


duke-nyuki:
Task linked: [QF-1806 Make sure the suggested project name is valid when creating a new project on QFIeldCloud](https://app.clickup.com/t/2192114/QF-1806)

nirvn (approved):

PR #418, opened 2022-04-15
Make sure the configuration made using project properties -> layer properties is not lost

### LINKED ISSUE #413: Configurable picture path not saved
Image path settings are not saved when we modify them and exit and reopen the current project settings.

image

Modify...
image

When I open properties again it saves the original route, not the modified one

suricactus:
When user opens layer properties via project properties and click Apply/Ok on both dialog windows, the layer settings (layer action or image path) get lost as they are cached in the layer source of the project properties window. This PR aims to fix this.

There are a few extra CPU cycles involved as the project "setdirty" signal is triggered multiple times.

Fix #413


duke-nyuki:
Task linked: [QF-1917 Configurable picture path not saved](https://app.clickup.com/t/2192114/QF-1917)

m-kuhn (approved):

PR #420, opened 2022-04-15
Orthographic fix on text strings

ValPinnaSardinia:
Fixing a few orthographic errors spotted during translation.


suricactus (approved):
Thanks!

PR #423, opened 2022-04-27
Prevent python error popping up because of deleted QgsMapLayer

suricactus:
Unfortunately there is no easy way to understand when the layer properties dialog was closed. On the other hand, depending on the `setDirty` event does not guarantee the wrapped C++ layer objects are not deleted.

Wrapping the code in try/except at least prevents the error popping up to the user until a better API is available.

### DIFF
--- a/qfieldsync/gui/layers_config_widget.py
+++ b/qfieldsync/gui/layers_config_widget.py
@@ -94,6 +94,10 @@ class LayersConfigWidget(QWidget, LayersConfigWidgetUi):
 
         self.reloadProject()
 
+    def closeEvent(self, event):
+        if Qgis.QGIS_VERSION_INT >= 31900:
+            self.project.dirtySet.disconnect(self._on_dirtyset)
+
     def get_available_actions(self, layer_source):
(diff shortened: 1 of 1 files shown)


duke-nyuki:
Task linked: [QF-1929 Release QFieldSync 4.1](https://app.clickup.com/t/2192114/QF-1929)

qfieldsync/gui/layers_config_widget.py
```diff
@@ -248,10 +252,19 @@ def toggleMenu_triggered(self, action):
...
-                layer_source.read_layer()
-
-            self.reloadProject()
+            # the layer might got deleted by the time dirtyset is called
+            try:
+                for layer_source in self.layer_sources:
+                    layer_source.read_layer()
+
+                self.reloadProject()
+            except Exception:
```
m-kuhn, line 261: Can we use a more specific exception here?

suricactus: `RuntimeError`, better than nothing

PR #449, opened 2022-08-02
Add "Attachement Directories" configuration on QFieldSync

suricactus:
So far QFieldSync automatically syncs only the `DCIM` directory. Now this is the default behaviour, but the user can set a list of directories that can be copied in the package too.

image


duke-nyuki:
Task linked: [QF-1252 Add "Attachement Directories" configuration on QFieldSync](https://app.clickup.com/t/2192114/QF-1252)

PR #453, opened 2022-08-09
Cut the response by the content length, so we can discard other bytes coming from buggy clients

suricactus:
Workaround to https://github.com/qgis/QGIS/issues/49687


duke-nyuki:
Task linked: [QF-2081 QFieldSync fix how errors are shown in the UI when 502 and refresh the projects list](https://app.clickup.com/t/2192114/QF-2081)

m-kuhn (approved):
Thanks

PR #455, opened 2022-08-10
Improved messages when the API returns an error

suricactus:


duke-nyuki:
Task linked: [QF-2089 Imrpove error messages when API returns an error](https://app.clickup.com/t/2192114/QF-2089)

PR #456, opened 2022-08-10
When there is a warning when creating a new project, just accept it

suricactus:


duke-nyuki:
Task linked: [QF-2068 create empty project out of QGIS is not allowed without an qgs/qgz file](https://app.clickup.com/t/2192114/QF-2068)

PR #457, opened 2022-08-10
Improve labels when loading a project and showing an error

suricactus:


duke-nyuki:
Task linked: [QF-1983 When loading the project list, inform the user that we are actually busy, now the list remains empty for N secs](https://app.clickup.com/t/2192114/QF-1983)

PR #479, opened 2023-01-28
Move away from photo-centric external resource support into file,image,audio,video

nirvn:
See https://github.com/opengisch/QField/pull/3851 on the QField side and the recent work done in QGIS itself such as https://github.com/qgis/QGIS/pull/51594.

Relevant libqfieldsync PR: https://github.com/opengisch/libqfieldsync/pull/41

UI slightly changed to add a tips label:
image

### DIFF
--- a/qfieldsync/gui/attachment_naming_widget.py
+++ b/qfieldsync/gui/attachment_naming_widget.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """
 /***************************************************************************
- ResourceNamingWidget
+ AttachmentNamingTableWidget
                                  A QGIS plugin
  Sync your projects to QField on android
                              -------------------
@@ -27,9 +27,9 @@ from qgis.PyQt.QtCore import Qt
 from qgis.PyQt.QtWidgets import QTableWidget, QTableWidgetItem
 
 
-class ResourceNamingTableWidget(QTableWidget):
+class AttachmentNamingTableWidget(QTableWidget):
     def __init__(self):
-        super(ResourceNamingTableWidget, self).__init__()
+        super(AttachmentNamingTableWidget, self).__init__()
 
         self.setColumnCount(3)
         self.setHorizontalHeaderLabels(
@@ -54,7 +54,6 @@ class ResourceNamingTableWidget(QTableWidget):
             ews = layer.editorWidgetSetup(i)
 
             if ews.type() == "ExternalResource":
-                # for later: if ews.config().get('DocumentViewer', QgsExternalResourceWidget.NoContent) == QgsExternalResourceWidget.Image:
                 self.insertRow(row)
                 item = QTableWidgetItem(layer.name())
                 item.setData(Qt.UserRole, layer_source)
@@ -70,7 +69,7 @@ class ResourceNamingTableWidget(QTableWidget):
                     if "DocumentViewer" in ews.config()
                     else 0
                 )
-                expression = layer_source.resource_naming(field.name(), resource_type)
+                expression = layer_source.attachment_naming(field.name(), resource_type)
                 ew.setExpression(expression)
                 self.setCellWidget(row, 2, ew)
 
@@ -84,7 +83,7 @@ class ResourceNamingTableWidget(QTableWidget):
             layer_source = self.item(i, 0).data(Qt.UserRole)
             field_name = self.item(i, 1).text()
             new_expression = self.cellWidget(i, 2).currentText()
-            layer_source.set_resource_naming(field_name, new_expression)
+            layer_source.set_attachment_naming(field_name, new_expression)
 
             if should_apply:
                 layer_source.apply()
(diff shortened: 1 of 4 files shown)


qfieldsync/gui/photo_naming_widget.py
```diff
@@ -52,7 +52,6 @@ def addLayerFields(self, layer_source):
         for i, field in enumerate(layer.fields()):
             row = self.rowCount()
             ews = layer.editorWidgetSetup(i)
-
             if ews.type() == "ExternalResource":
```
suricactus, line 55: can we reduce diff and keep the nice separation between vars and ifs?

nirvn: @suricactus , done -- sorry, I pushed a wrong commit earlier, some more things to review.

suricactus:
Same comment as in lib. Iwould prefer we stick to "attachments" only.

nirvn:
@suricactus , see the comment here: https://github.com/opengisch/libqfieldsync/pull/41#issuecomment-1407374898

qfieldsync/gui/map_layer_config_widget.py
```diff
@@ -74,14 +75,23 @@ def __init__(self, layer, canvas, parent):
...
+        # append the attachment naming table to the layout
         if layer.type() == QgsMapLayer.VectorLayer:
-            self.layout().insertRow(-1, self.tr("Photo naming"), self.photoNamingTable)
-            self.photoNamingTable.setEnabled(self.photoNamingTable.rowCount() > 0)
+            self.layout().insertRow(
+                -1, self.tr("Attachment\nnaming"), self.attachmentNamingTable
+            )
+            tip = QLabel(
+                self.tr(
+                    "In your expressions, use #~filename~# and #~extension~# tags to refer to attachement file name and extension."
```
suricactus, line 89: wait on the template placeholders to confirm the syntax.

qfieldsync/gui/map_layer_config_widget.py
```diff
@@ -91,7 +101,7 @@ def __init__(self, layer, canvas, parent):
         if layer.type() == QgsMapLayer.VectorLayer:
             self.layout().insertRow(
                 -1,
-                self.tr("Relationship configuration"),
+                self.tr("Relationship\nconfiguration"),
```
suricactus, line 104: Why don't you `QLabel.setWordWrap(true);`, instead of forcing newline? Or QT is not smart enough and makes the first column too wide?

qfieldsync/gui/map_layer_config_widget.py
```diff
@@ -74,14 +75,23 @@ def __init__(self, layer, canvas, parent):
...
+        self.attachmentNamingTable.addLayerFields(self.layer_source)
+        self.attachmentNamingTable.setLayerColumnHidden(True)
 
-        # append the photo naming table to the layout
+        # append the attachment naming table to the layout
         if layer.type() == QgsMapLayer.VectorLayer:
-            self.layout().insertRow(-1, self.tr("Photo naming"), self.photoNamingTable)
-            self.photoNamingTable.setEnabled(self.photoNamingTable.rowCount() > 0)
+            self.layout().insertRow(
+                -1, self.tr("Attachment\nnaming"), self.attachmentNamingTable
```
suricactus, line 85: Why don't you `QLabel.setWordWrap(true);`, instead of forcing newline? Or QT is not smart enough and makes the first column too wide?

qfieldsync/gui/attachment_naming_widget.py
```diff
@@ -65,7 +64,12 @@ def addLayerFields(self, layer_source):
                 self.setItem(row, 1, item)
                 ew = QgsFieldExpressionWidget()
                 ew.setLayer(layer)
-                expression = layer_source.photo_naming(field.name())
+                resource_type = (
+                    ews.config()["DocumentViewer"]
+                    if "DocumentViewer" in ews.config()
+                    else 0
+                )
+                expression = layer_source.attachment_naming(field.name(), resource_type)
```
suricactus, line 72: See my comment here: https://github.com/opengisch/libqfieldsync/pull/41#discussion_r1089725178

nirvn: @suricactus , it shouldn't impact this line. Here the value is the QgsExternalResourceWidget.DocumentViewerContent enum values:

https://github.com/qgis/QGIS/blob/master/src/gui/qgsexternalresourcewidget.h#L74

If libqfieldsync has a qgis gui dependence already, we can use the enum there too.

suricactus: Ah, but it should be in qgis.gui, no? It's data stored in the project file, not GUI specific enum... IMO we should have our own `Enum` copying `QgsExternalResourceWidget.DocumentViewerContent` in libqfieldsync and adding a comment why we replicated it. It would be super handy also for the server code in QFC, since QGIS container there is built without the `gui` library at all.

qfieldsync/gui/map_layer_config_widget.py
nirvn: Not smart enough

qfieldsync/gui/map_layer_config_widget.py
nirvn: Not smart enough

nirvn:
@suricactus , address your comments.

PR #480, opened 2023-02-03
Increase the timeout for all requests to QFC

suricactus:
Whenever QFC API is requested to list the project files, it asks the S3 API to return the list of files and versions. As the file list grows, it only gets slower. In their case it takes 1.5min+ to respond for 4k files.

1) Fastest: make the API respond with files only, so don't iterate 10*N files, but N only.
- server side changes, around 0.5-1 day
- qfieldsync changes, so it calls the new endpoint that serves the files.

2) Medium: store all file metadata in the DB, instead of polling the S3 storage. DB should be way faster.
- server side only, few days

3) Slow: we need to add pagination on pretty much all data.
- server side, few days
- qfieldsynnc
- qfield
- qfieldcloud-sdk
-

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -37,7 +37,7 @@ jobs:
 
   # Run unit tests
   test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


signedav (approved):
As far as I see it in the context.

PR #481, opened 2023-02-11
Avoid crashing when the importDirectoryProject preference is not a valid JSON

nirvn:
This should fix https://github.com/opengisch/QField/issues/3896 -- not sure why 4.3.0 would have triggered the regression since none of the code / preference involved here is new. In any case, this commit harden the code.

### DIFF
--- a/qfieldsync/gui/synchronize_dialog.py
+++ b/qfieldsync/gui/synchronize_dialog.py
@@ -65,7 +65,10 @@ class SynchronizeDialog(QDialog, DialogUi):
         self.button_box.button(QDialogButtonBox.Save).clicked.connect(
             lambda: self.start_synchronization()
         )
-        import_dir = self.preferences.value("importDirectoryProject")
+        try:
+            import_dir = self.preferences.value("importDirectoryProject")
+        except ValueError:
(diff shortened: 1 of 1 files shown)


qfieldsync/gui/synchronize_dialog.py
```diff
@@ -65,7 +65,10 @@ def __init__(self, iface, offline_editing, parent=None):
         self.button_box.button(QDialogButtonBox.Save).clicked.connect(
             lambda: self.start_synchronization()
         )
-        import_dir = self.preferences.value("importDirectoryProject")
+        try:
+            import_dir = self.preferences.value("importDirectoryProject")
+        except ValueError:
+            import_dir = None
         if not import_dir:
             import_dir = self.preferences.value("importDirectory")
```
m-kuhn, line 73: Does this line also need to be hardened?

m-kuhn (approved):

qfieldsync/gui/synchronize_dialog.py
nirvn: @m-kuhn , no, that setting isn't a dictionary.

PR #485, opened 2023-02-13
Bump min qgis version to 3.22

m-kuhn:


suricactus (approved):
Good move. But I am pretty sure we do work on 3.16.13+ (the python version upgrade) on windows.

m-kuhn:
Hmm... Can we meanwhile have minimum on patch version?

suricactus:
We already tried, but plugins.qgis.org and QGIS did not support it back then. There was development done on this front, but I never tested if it's deployed. I cannot test this myself now.

m-kuhn:
We should probably also change this anyway

https://github.com/opengisch/qfieldsync/blob/73c323457af141b70d886c07950a2da813d84211/.github/workflows/continuous_integration.yml#L43

suricactus:
Actually you just confirmed that QFieldSync works on newer QGIS 3.16.latest version.

m-kuhn:
I think many organisations lag a bit behind but one year after retiring 3.16 officially, it's time to move on for us.

suricactus (approved):

PR #526, opened 2023-06-17
Package libqfieldsync as wheel

m-kuhn:
Depends on
- https://github.com/opengisch/qgis-plugin-ci/pull/230
- https://github.com/opengisch/libqfieldsync/pull/49

Developer information, to work locally, either install the `requirements.txt` (if you don't need to edit libqfieldsync) or pip install libqfieldsync as editable if you need to work on libqfieldsync.

```sh
git clone git@github.com:opengisch/libqfieldsync.git
pip install -e libqfieldsync
```

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -71,8 +71,7 @@ jobs:
           sudo pip3 install --upgrade pip
           sudo pip3 install pytest nose2 mock
 
-          pip3 wheel $(grep -o -P '(https://.*.tar.gz)' requirements.txt)
-          pip3 install libqfieldsync-*.whl
+          pip3 install -r requirements.txt
 
           xvfb-run pytest
 
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -41,7 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        qgis_version: [release-3_22, release-3_28, latest]
+        qgis_version: [release-3_22, latest]
     env:
       QGIS_TEST_VERSION: ${{ matrix.qgis_version }}
     steps:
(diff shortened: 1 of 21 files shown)


requirements.txt
```diff
@@ -4,3 +4,6 @@ flake8-respect-noqa
 pytest
 future
 transifex-client
+
+# TODO: point to a commit on master after merging https://github.com/opengisch/libqfieldsync/pull/49
```
m-kuhn, line 8: Open todo ...

.github/workflows/continuous_integration.yml
```diff
@@ -89,6 +90,10 @@ jobs:
         run: |
           sudo apt update && sudo apt install qtbase5-dev qttools5-dev-tools
           sudo pip install qgis-plugin-ci
+      - name: Package libqfieldsync
+        run: |
+          pip wheel $(grep -o -P '(https://.*.tar.gz)' requirements.txt)
```
suricactus, line 95:
```suggestion
          pip wheel $(grep -o -P '(https://.*\.tar\.gz)' requirements.txt)
```

qfieldsync/__init__.py
```diff
@@ -25,9 +25,17 @@
...
 
 import os
+import pathlib
+import sys
 
 import qgis.utils
 
+src_dir = pathlib.Path(__file__).parent.resolve()
+
+libqfieldsync_whl = src_dir / "libqfieldsync.whl"
```
suricactus, line 35:
```suggestion
libqfieldsync_whl = src_dir.joinpath("libqfieldsync.whl")
```
I always feel afraid of too much magic with operator overload.

m-kuhn:
@suricactus this should be ready.
One TODO: point to a commit on master after merging the PR on https://github.com/opengisch/libqfieldsync/pull/49

m-kuhn:
I think we will want to points to a sha on master and update explicitly

suricactus:
Yes, we want that for releases, I do have it mind. Will pinpoint when I am about to release a new version, there is a bit of other development to be made for now.

Also thinking whether to create a `release` branch as I did on QFieldCloud and pinpoint only there, so we are forced to keep libqfieldsync and qfieldsync master branches compatible with each other, until we think it's not working.

PR #529, opened 2023-07-21
Fix copying file dirs to skip the `.qfieldsync` directory

suricactus:


qfieldsync/utils/file_utils.py
```diff
@@ -51,6 +51,9 @@ def path_to_dict(path: PathLike, dirs_only: bool = False) -> DirectoryTreeDict:
             if dirs_only and not subpath.is_dir():
                 continue
 
+            if "/.qfieldsync" in str(subpath):
```
m-kuhn, line 54: The initial `/` is on purpose?

suricactus: Doesn't really matter, this is slightly stricter. Will force change it without the slash so it does not raise questions anymore.

PR #532, opened 2023-07-21
Show server side error when login failed

### LINKED ISSUE #511: Sign in error "None"
The error message "None" is not very descriptif

suricactus:
EDIT:
Even better now.
image

---

So far it was showing "Failed to login: None"

image

Fixes #511

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -682,14 +682,16 @@ class CloudNetworkAccessManager(QObject):
 
         error = self.user_details.get("error")
         if self._login_error:
-            if self._login_error.httpCode is None:
+            if self._login_error.httpCode is None or (
+                self._login_error.httpCode >= 400 and self._login_error.httpCode < 500
+            ):
                 error = str(self._login_error)
             elif self._login_error.httpCode >= 500:
                 error = self.tr("Server error {}").format(self._login_error.httpCode)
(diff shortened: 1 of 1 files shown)


qfieldsync/core/cloud_api.py
```diff
@@ -682,14 +682,16 @@ def get_last_login_error(self) -> str:
 
         error = self.user_details.get("error")
         if self._login_error:
-            if self._login_error.httpCode is None:
+            if self._login_error.httpCode is None or (
+                self._login_error.httpCode >= 400 and self._login_error.httpCode < 500
+            ):
                 error = str(self._login_error)
             elif self._login_error.httpCode >= 500:
                 error = self.tr("Server error {}").format(self._login_error.httpCode)
```
faebebin, line 690:
```suggestion
            http_code = self._login_error.httpCode
            if http_code and http_code >= 500:
                error = self.tr("Server error {}").format(http_code)
            elif http_code is None or http_code >= 400:
                error = str(self._login_error)
```

faebebin: just some idea, not sure bout correctness But creating a variable would improve readability anyway?

suricactus: Agreed. I am not a big fan of reassigning values to new variables, but in this case it is better. Thanks!

PR #534, opened 2023-07-21
Make it possible to select project owner during project creation

suricactus:
| before | after |
|-|-|
| image | image |

### DIFF
--- a/qfieldsync/gui/cloud_create_project_widget.py
+++ b/qfieldsync/gui/cloud_create_project_widget.py
@@ -127,6 +127,10 @@ class CloudCreateProjectWidget(QWidget, WidgetUi):
             )
         )
 
+        self.projectOwnerRefreshButton.clicked.connect(
+            lambda: self.on_project_owner_refresh_button_click()
+        )
+
     def restart(self):
         self.stackedWidget.setCurrentWidget(self.selectTypePage)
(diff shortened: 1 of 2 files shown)


duke-nyuki:
Task linked: [QF-2863 Make it possible to select project owner during project creation](https://app.clickup.com/t/2192114/QF-2863)

qfieldsync/gui/cloud_create_project_widget.py
```diff
@@ -127,6 +127,10 @@ def __init__(
             )
         )
 
+        self.projectOwnerRefreshButton.clicked.connect(
+            lambda: self.on_project_owner_refresh_button_click()
+        )
```
faebebin, line 132: this necessary? can't it just be `connect(self.on_project_owner_refresh_button_click)` ?

suricactus: Nope, you are missing the `self` in `on_project_owner_refresh_button_click` then. Theoretically you can define it as a class method, but it's ugly and in case you need the state in `on_project_owner_refresh_button_click`, you still need to have `self`.

faebebin: thanks for info :+1:

PR #543, opened 2023-11-01
Give full with to login dialog widgets

m-kuhn:
- Remember me checkbox
- Error messages
- New user registration

Before
image
After
image


nirvn:
@m-kuhn , nice set of improvements, is there a way we could harmonize the margins (i.e. for the user and password label not to having a bigger left/right margin?

m-kuhn:
> @m-kuhn , nice set of improvements, is there a way we could harmonize the margins (i.e. for the user and password label not to having a bigger left/right margin?

Thanks, good idea, left and right margin are 0 now

nirvn (approved):
Thanks

PR #544, opened 2023-11-18
Insure login dialog is front and center

nirvn:
Calling raise() and activateWindow() is used to insure dialogs are raised to the top, let's call this to avoid dialog stacking confusion on macos.

Also, the cloud projects dialog was mishandling its visible state (calling self.hide() prior to opening the login window only to have it re-shown after constructor (by show_cloud_overview_dialog(self)). This lead to the (disabled) projects dialog being focused instead of the login dialog. I suspect that will also alleviate whatever is going on with macos.

### DIFF
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -121,20 +121,6 @@ class CloudProjectsDialog(QDialog, CloudProjectsDialogUi):
 
         self.update_welcome_label()
 
-        if self.network_manager.has_token():
-            self.show_projects()
-            self.show()
-            self.createButton.setEnabled(True)
-        else:
-            CloudLoginDialog.show_auth_dialog(
-                self.network_manager,
-                lambda: self.on_auth_accepted(),
-                lambda: self.close(),
-                parent=self,
-            )
-            self.hide()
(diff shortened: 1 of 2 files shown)


qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -121,20 +121,6 @@ def __init__(
...
-        else:
-            CloudLoginDialog.show_auth_dialog(
-                self.network_manager,
-                lambda: self.on_auth_accepted(),
-                lambda: self.close(),
-                parent=self,
-            )
-            self.hide()
-            self.createButton.setEnabled(False)
-
```
suricactus, line 137: Is it required to move this block more to the bottom?

suricactus (commented):
Thanks for dealing with this <3

qfieldsync/gui/cloud_projects_dialog.py
nirvn: @suricactus , I wanted to call self.show() after as much of the UI was set up as possible. It however has to be before we create CloudCreateProjectWidget has that will also open the login dialog within the constructor.

suricactus:
Works perfect here! Thanks!

PR #546, opened 2023-11-20
Update README.md with development instruction

suricactus:
Seems what has been described in https://github.com/opengisch/qfieldsync/pull/526 was never shared in the docs.


README.md
```diff
@@ -10,3 +10,30 @@ It analyses the QGIS project and suggests and performs actions needed to make th
...
+3) Checkout [libqfieldsync](https://github.com/opengisch/libqfieldsync/) locally:
+
+```
+git clone git@github.com:opengisch/libqfieldsync.git
+```
+
+4) Install your local libqfieldsync as editable dependency (assuming you are in the same directory as step 3):
+
+```
+pip install -e libqfieldsync
```
nirvn, line 37: on ubuntu 23.10 (and on latest debian too I think), you'll need to add a scary --break-system-packages to that line.

suricactus: Accomodated.

PR #548, opened 2023-12-01
Implement a feature-based geometry lock mechanism

nirvn:
This PR implements a functionality in qfieldsync allowing users to determine the locked geometry state of individual features within a vector layer.

First, here's a screenshot of the UI:

image

A brand new data-defined property button sits next to the [x] lock geometry checkbox. From there, users can select a feature attribute or build a more complex expression that will change the locked geometry status of individual features.

_(Requires https://github.com/opengisch/libqfieldsync/pull/58 on the libqfieldsync side of things)_

### DIFF
--- a/qfieldsync/gui/map_layer_config_widget.py
+++ b/qfieldsync/gui/map_layer_config_widget.py
@@ -23,7 +23,7 @@
 import os
 
 from libqfieldsync.layer import LayerSource
-from qgis.core import QgsMapLayer, QgsProject
+from qgis.core import QgsMapLayer, QgsProject, QgsProperty, QgsPropertyDefinition
 from qgis.gui import QgsMapLayerConfigWidget, QgsMapLayerConfigWidgetFactory
 from qgis.PyQt.QtWidgets import QLabel
 from qgis.PyQt.uic import loadUiType
@@ -55,6 +55,8 @@ class MapLayerConfigWidgetFactory(QgsMapLayerConfigWidgetFactory):
 
 
 class MapLayerConfigWidget(QgsMapLayerConfigWidget, WidgetUi):
+    PROPERTY_GEOMETRY_LOCKED = 1
+
     def __init__(self, layer, canvas, parent):
         super(MapLayerConfigWidget, self).__init__(layer, canvas, parent)
         self.setupUi(self)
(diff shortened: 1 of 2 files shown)


qfieldsync/ui/map_layer_config_widget.ui
```diff
@@ -39,20 +39,51 @@
...
+       <property name="toolTip">
+        <string>When enabled, this option disables adding and deleting features, as well as modifying the geometries of existing features.</string>
+       </property>
+       <property name="text">
+        <string>Lock geometries</string>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QgsPropertyOverrideButton" name="isGeometryLockedDDButton">
```
suricactus, line 54: What DD stands for?

qfieldsync/gui/map_layer_config_widget.py
```diff
@@ -55,6 +55,8 @@ def supportLayerPropertiesDialog(self):
 
 
 class MapLayerConfigWidget(QgsMapLayerConfigWidget, WidgetUi):
+    PROPERTY_GEOMETRY_LOCKED = 1
```
suricactus, line 58: It is not super clear to me what the `1` stands for?

suricactus (approved):
Two comments on the code, just clarifications on naming. Very nice it is using the same mechanism to define dynamic value as in the rest of QGIS.

qfieldsync/gui/map_layer_config_widget.py
nirvn: @suricactus , with  properties, you need to provide an integer identifier for a given property (see https://github.com/opengisch/qfieldsync/pull/548/files#diff-cb0fe03aac5d043fa6713776d09aaa241751d0e5b4f9d8d7c2ee1b20d58a6905R89), much better to have it as a PROPERTY_SOMETHING constant than an integer of 1 entered in the data-defined button's init() function.

qfieldsync/ui/map_layer_config_widget.ui
suricactus: Thanks!

qfieldsync/gui/map_layer_config_widget.py
suricactus: I see. Thanks for the clarification!

suricactus (approved):

PR #557, opened 2024-01-08
Use the new `QgisCoreOffliner` that wraps `QgsOfflineEditing`

suricactus:


requirements.txt
```diff
@@ -5,4 +5,4 @@ pytest
 future
 transifex-client
 
-libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/a6162194f5b7e2448da26d3cf67ed542653c4012.tar.gz
+libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/a8e9a3a4a341f797934e76c747938766362c462c.tar.gz
```
m-kuhn, line 8: This points to a PR branch, correct?

I would always point that to something on master, in case of force-push, branch-delete-after-merge or what-do-i-know bad things may happen.

PR #567, opened 2024-02-29
Tracking session UI

nirvn:
I'll let the screenshot do the talking :)

image

(requires libqfieldsync' s https://github.com/opengisch/libqfieldsync/pull/71)


suricactus (approved):
Code looks good! Nice improvement. Having old/new screenshot would have made the contrast even bigger.

PR #584, opened 2024-05-31
Implement UI to control the new value map editor widget's toggle buttons interface

nirvn:
In the future, it'd be great to have the UX implemented in QGIS itself. But for now, let's rely on our beloved qfieldsync to implement a simple layer-level configuration:

image

### DIFF
--- a/qfieldsync/ui/map_layer_config_widget.ui
+++ b/qfieldsync/ui/map_layer_config_widget.ui
@@ -87,9 +87,57 @@
    <item>
     <widget class="QGroupBox" name="attachmentsRelationsGroupBox">
      <property name="title">
-      <string>Attachments and Relationships Settings</string>
+      <string>Feature Form, Attachments and Relationships Settings</string>
      </property>
      <layout class="QFormLayout" name="attachmentsRelationsLayout">
+      <item row="0" column="0" colspan="2">
+       <layout class="QHBoxLayout" name="valueMapButtonInterfaceLayout">
+        <item>
+         <widget class="QLabel" name="valueMapButtonInterfaceLabel">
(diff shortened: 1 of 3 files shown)


domi4484:
Why do we do that per layer and not per field? too fine graded?

nirvn:
@domi4484 , yeah, I don't think we have the UI in place to do per-field, and I'm hopeful this will be a temporary setting in any case.

qfieldsync/ui/map_layer_config_widget.ui
```diff
@@ -87,9 +87,57 @@
...
+        <item>
+         <widget class="QLabel" name="valueMapButtonInterfaceLabel">
+          <property name="sizePolicy">
+           <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+            <horstretch>1</horstretch>
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+          <property name="text">
+           <string>Item threshold under which value map editor widgets will use a toggle buttons interface</string>
```
suricactus, line 104:
```suggestion
           <string>Item threshold under which value map editor widgets will use toggle buttons interface</string>
```

I guess the "a" is incompatible with the plural "buttons"?

suricactus (approved):

qfieldsync/ui/map_layer_config_widget.ui
nirvn: I think a works here as it is a[n] interface.

suricactus: Ah, my poor English. Good!

PR #593, opened 2024-06-29
Implement geofencing settings in the project properties' QField panel

nirvn:
This PR adds a couple of widgets to setup geofencing in QField. Looks:

image

The libqfieldsync counterpart is here: https://github.com/opengisch/libqfieldsync/pull/86

### DIFF
--- a/qfieldsync/gui/project_configuration_widget.py
+++ b/qfieldsync/gui/project_configuration_widget.py
@@ -110,6 +110,19 @@ class ProjectConfigurationWidget(WidgetUi, QgsOptionsPageWidget):
         self.event_eater = EventEater()
         self.attachmentDirsListWidget.installEventFilter(self.event_eater)
 
+        self.geofencingBehaviorComboBox.addItem(
+            self.tr("Alert users when inside an area"),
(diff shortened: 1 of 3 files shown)


m-kuhn:
Nice !

Can we have a UI adjustment: the description what it does (i.e. it will alert in case of XXX) should be there. The "reverted logic" should live as a tooltip instead.

nirvn:
@m-kuhn , we can certainly have a UI adjustment ;) good idea.

qfieldsync/gui/project_configuration_widget.py
```diff
@@ -229,21 +269,56 @@ def apply(self):
...
             )
         except AttributeError:
             pass
+
+        # Geofencing settings
+        self.__project_configuration.geofencing_is_active = (
+            self.geofencingGroupBox.isChecked()
+        )
+
+        try:
```
suricactus, line 304: Do you have a concrete example when we need the `try/except`? I see it is also 10 lines above, but on this line it seems you handle this with the `if/else` check for `currentLayer()` existance.

suricactus (approved):
Looks good to me, one small comment.

PR #601, opened 2024-07-21
Add map themes's default active layer configuration

nirvn:
This PR implements the project configuration UI to setup the default active layer QField will pick when users select a map theme in QField. The functionality allows for project managers to help out field workers by auto-picking the right active layer to digitize entries based on map themes.

Screenshot:

image


suricactus:
This configuration panel is getting super busy lately. We can also move this setting to the individual layers, but the overview is indeed better. No need to address this now, just sharing. Thanks for the contribution!

nirvn:
@suricactus , yeah, I had the same feeling when developing both geofencing and this feature.

Issue #605, opened 2024-08-19
Bug: QField Sync Plugin Removes Path Segment from URL


edouardbruelhart:
I've encountered a bug with the QField Sync plugin where it incorrectly handles the base URL for connecting to a self-hosted QFieldCloud instance, served using a custom nginx configuration. Specifically, the plugin is removing the /qfieldcloud/ segment from the URL, which leads to a 404 Not Found error when attempting to authenticate or interact with the API.

**Steps to Reproduce:**

  - Configure the QField Sync plugin with a base URL that includes the /qfieldcloud/ segment, e.g., https://emi-collection.unifr.ch/qfieldcloud/.
  - Attempt to perform an action that requires authentication or interaction with the API (e.g., logging in or syncing).
  - Observe that the plugin makes requests to URLs without the /qfieldcloud/ segment, e.g., https://emi-collection.unifr.ch/api/v1/auth/login/.

**Expected Behavior:**

The QField Sync plugin should use the full base URL provided, including all path segments, so that requests are correctly routed to the API endpoints. For example, the URL https://emi-collection.unifr.ch/qfieldcloud/api/v1/auth/login/ should be used instead of https://emi-collection.unifr.ch/api/v1/auth/login/.

**Actual Behavior:**

The plugin removes the /qfieldcloud/ path segment from the URL, resulting in incorrect requests like https://emi-collection.unifr.ch/api/v1/auth/login/, which leads to 404 Not Found errors.

**Additional Information:**

  I have tested to connect to the instance using a web browser, QField application and directly using curl and confirmed that it works with the /qfieldcloud/ path segment.
  The issue seems to be specific to how the plugin constructs URLs.

**Environment:**

  - QField Sync Plugin Version: 4.10.1
  - QGIS Version: 3.38.2
  - Operating System: Ubuntu 22.04
  - QFieldCloud URL: https://emi-collection.unifr.ch/qfieldcloud/

Thank you for your attention to this issue. Please let me know if you need any further details or if there is a workaround available.

m-kuhn:
Thank you for reporting this, we will happily accept a pull request to improve this !

edouardbruelhart:
I added a pull request to correct this. Let me know if it suits for you or if you want another mechanism to perform this

PR #607, opened 2024-08-20
correction poposition for issue 605

edouardbruelhart:
Here I propose to ignore paths that start with "/api" or that are equal to "/" to remove unnecessary parts, but to keep other ones to avoid 404 errors.


suricactus:
This PR fixes https://github.com/opengisch/qfieldsync/issues/605#issuecomment-2298187889

suricactus (approved):
Thanks for your contribution! It seems I never published my review... 

Happy to merge once we fix the comments!

suricactus (changes requested):
Ah, also if you can run `pre-commit` to make the code checks happy.

suricactus (approved):
Thanks!

PR #614, opened 2024-09-21
Adding validator for files on converter to cloud project

SeqLaz:
Based on: https://github.com/opengisch/libqfieldsync/pull/95

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,4 @@ future
 transifex-client
 
 # NOTE `libqfielsync` version should be defined in the `*.tar.gz` format, not `git+https://` to make `wheel` happy
-libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/7901b65225caa2fb8788639f1e21b1050c87a6f3.tar.gz
+libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/5ba992c1f91ae74418a07c372a7ae11db557b10e.tar.gz


suricactus (commented):
I would add such function in file utils.

Also the pattern is a bit more complex, expecially for windows. Can we use something like this:
```
\A(?!(?:COM[0-9]|CON|LPT[0-9]|NUL|PRN|AUX|com[0-9]|con|lpt[0-9]|nul|prn|aux)|\s|[\.]{2,})[^\\\/:*"?<>|]{1,254}(?<![\s\.])\z
```

Even better, we can check if `pathvalidate` is installed and use that, otherwise fallback to the above regex (or better one). Don't forget to check if the file path is also correct.

SeqLaz:
> I would add such function in file utils.
> 
> Also the pattern is a bit more complex, expecially for windows. Can we use something like this:
> 
> ```
> \A(?!(?:COM[0-9]|CON|LPT[0-9]|NUL|PRN|AUX|com[0-9]|con|lpt[0-9]|nul|prn|aux)|\s|[\.]{2,})[^\\\/:*"?<>|]{1,254}(?<![\s\.])\z
> ```
> 
> Even better, we can check if `pathvalidate` is installed and use that, otherwise fallback to the above regex (or better one). Don't forget to check if the file path is also correct.

Thank you. Due to the library `pathvalidate` not being installed, I used your regex for the validations and moved the function to `/utils/file_utils.py`.

qfieldsync/core/cloud_converter.py
```diff
@@ -128,6 +129,19 @@ def convert(self) -> None:  # noqa: C901
                             self.project.removeMapLayer(layer)
                             continue
                 else:
+                    # Validate filenames and paths before copying
+                    if not is_valid_filename(layer.name()) or not is_valid_path(
+                        layer.source()
+                    ):
```
suricactus, line 135:
```suggestion
                    if not is_valid_path(str(layer.filename)):
```
I am not sure if we need to run `str`, but the idea is that `is_valid_path` is already using `is_valid_filename`

qfieldsync/utils/file_utils.py
```diff
@@ -61,3 +62,27 @@ def path_to_dict(path: PathLike, dirs_only: bool = False) -> DirectoryTreeDict:
...
+    """
+    Check if the filename is valid.
+    """
+    pattern = re.compile(
+        r'\A(?!(?:COM[0-9]|CON|LPT[0-9]|NUL|PRN|AUX|com[0-9]|con|lpt[0-9]|nul|prn|aux)|\s|[\.]{2,})[^\\\/:*"?<>|]{1,254}(?<![\s\.])\z'
+    )
+    return bool(pattern.match(filename))
+
+
+def is_valid_path(path: str) -> bool:
```
suricactus, line 77:
```suggestion
def is_valid_filepath(path: str) -> bool:
```

Can we use the same name as in the quoted library?

suricactus (commented):
Can we use the `ProjectChecker` for providing this kind of feedback? There is already ASCII filename warnings, the file one to be added should be an error that prevents further actions with the project.

qfieldsync/core/cloud_converter.py
SeqLaz: You are right, completely agree in the path validation have validation for the filename!

qfieldsync/utils/file_utils.py
SeqLaz: Yes for sure, is way more better!

SeqLaz:
> `ProjectChecker`



> Can we use the `ProjectChecker` for providing this kind of feedback? There is already ASCII filename warnings, the file one to be added should be an error that prevents further actions with the project.

For this it will be require to add a method to libqfieldsync

qfieldsync/utils/file_utils.py
SeqLaz: Great!

suricactus:
Can you add a screencast how this works? Code looks good enough for now.

SeqLaz:
> Can you add a screencast how this works? Code looks good enough for now.

There in the left is with when the message is directly on QFieldSync on the right when is using the libqfieldsync

| Before | After |
|--------|-------|
| https://github.com/user-attachments/assets/fbb89cdd-aa5e-4f2c-ab3a-758dd0119d1f | https://github.com/user-attachments/assets/875183e9-9658-4d21-bbe6-707657bd5691 |

My vote is for opt for the libqfieldsync option

suricactus (changes requested):
I like the second option too! Thanks, good work! Please link the libqfieldsync PR here when ready.

SeqLaz:
> I like the second option too! Thanks, good work! Please link the libqfieldsync PR here when ready.

If we opt for the second option we can close this PR, all the logic is managed in this PR https://github.com/opengisch/libqfieldsync/pull/95

suricactus:
Actually you can reopen this and just bump your linqfieldsync commit version.

SeqLaz:
Hey @suricactus I removed all the previous commits and add the last master as part of the requirement, in the last master was merged the [PR](https://github.com/opengisch/libqfieldsync/pull/95#event-14843461397)

suricactus (approved):

PR #615, opened 2024-09-25
Adding the option for specify the project title on packaged project

SeqLaz:
This adds the option to specify a new project name and title in the exported project when packaging.

| Before | After |
|--------|-------|
| Before Image | After Image |

Related PR: https://github.com/opengisch/libqfieldsync/pull/94

### DIFF
--- a/qfieldsync/gui/package_dialog.py
+++ b/qfieldsync/gui/package_dialog.py
@@ -74,7 +74,7 @@ class PackageDialog(QDialog, DialogUi):
         self.__project_configuration = ProjectConfiguration(self.project)
         self.button_box.button(QDialogButtonBox.Save).setText(self.tr("Create"))
         self.button_box.button(QDialogButtonBox.Save).clicked.connect(
-            self.package_project
+            self.run_package_project
         )
         self.button_box.button(QDialogButtonBox.Reset).setText(
             self.tr("Configure current project...")
@@ -162,6 +162,20 @@ class PackageDialog(QDialog, DialogUi):
         self.nextButton.setVisible(False)
         self.button_box.setVisible(True)
         self.stackedWidget.setCurrentWidget(self.packagePage)
+
+    def run_package_project(self):
+        export_packaged_project = Path(self.packagedProjectFileWidget.filePath())
+
+        if export_packaged_project.suffix != ".qgs":
+            QMessageBox.critical(
+                self,
+                self.tr("Invalid File"),
+                self.tr('The filename must have a ".qgs" extension.'),
+            )
+            return
+
+        else:
+            self.package_project()
 
     def package_project(self):
         self.button_box.button(QDialogButtonBox.Save).setEnabled(False)
--- a/qfieldsync/gui/package_dialog.py
+++ b/qfieldsync/gui/package_dialog.py
@@ -21,6 +21,7 @@
  ***************************************************************************/
 """
 import os
+from pathlib import Path
 
 from libqfieldsync.layer import LayerSource
 from libqfieldsync.offline_converter import ExportType, OfflineConverter
@@ -54,8 +55,6 @@ from qfieldsync.gui.checker_feedback_table import CheckerFeedbackTable
 from qfieldsync.gui.dirs_to_copy_widget import DirsToCopyWidget
 from qfieldsync.gui.project_configuration_dialog import ProjectConfigurationDialog
 
-from ..utils.qt_utils import make_folder_selector
-
 DialogUi, _ = loadUiType(
     os.path.join(os.path.dirname(__file__), "../ui/package_dialog.ui")
 )
@@ -73,7 +72,6 @@ class PackageDialog(QDialog, DialogUi):
         self.qfield_preferences = Preferences()
         self.dirsToCopyWidget = DirsToCopyWidget()
         self.__project_configuration = ProjectConfiguration(self.project)
-        self.project_lbl.setText(get_project_title(self.project))
         self.button_box.button(QDialogButtonBox.Save).setText(self.tr("Create"))
         self.button_box.button(QDialogButtonBox.Save).clicked.connect(
             self.package_project
@@ -105,15 +103,13 @@ class PackageDialog(QDialog, DialogUi):
 
     def setup_gui(self):
         """Populate gui and connect signals of the push dialog"""
-        export_dirname = self.qfield_preferences.value("exportDirectoryProject")
-        if not export_dirname:
-            export_dirname = os.path.join(
-                self.qfield_preferences.value("exportDirectory"),
-                fileparts(QgsProject.instance().fileName())[1],
-            )
+        self.packagedProjectTitleLineEdit.setText(get_project_title(self.project))
+        self.packagedProjectFileWidget.setFilter("QGIS Project Files (*.qgs)")
+        self.packagedProjectFileWidget.setConfirmOverwrite(True)
+        self.packagedProjectFileWidget.setFilePath(
+            self.get_export_filename_suggestion()
+        )
 
-        self.manualDir.setText(QDir.toNativeSeparators(str(export_dirname)))
-        self.manualDir_btn.clicked.connect(make_folder_selector(self.manualDir))
         self.update_info_visibility()
 
         self.nextButton.clicked.connect(lambda: self.show_package_page())
@@ -140,10 +136,27 @@ class PackageDialog(QDialog, DialogUi):
         else:
             self.show_package_page()
 
-    def get_export_folder_from_dialog(self):
+    def get_export_folder(self):
         """Get the export folder according to the inputs in the selected"""
-        # manual
-        return self.manualDir.text()
+        export_dirname = self.qfield_preferences.value("exportDirectoryProject")
+        if not export_dirname:
+            export_dirname = os.path.join(
+                self.qfield_preferences.value("exportDirectory"),
+                fileparts(QgsProject.instance().fileName())[1],
+            )
+        return QDir.toNativeSeparators(str(export_dirname))
+
+    def get_export_filename_suggestion(self):
+        export_folder = Path(self.get_export_folder())
+        full_project_name_suggestion = export_folder.joinpath(
+            f"{self.project.baseName()}_qfield.qgs"
+        )
+        return str(full_project_name_suggestion)
+
+    def set_export_filename_suggested(self):
+        self.packagedProjectFileWidget.setFilePath(
+            self.get_export_filename_suggestion()
+        )
 
     def show_package_page(self):
         self.nextButton.setVisible(False)
@@ -153,7 +166,7 @@ class PackageDialog(QDialog, DialogUi):
     def package_project(self):
         self.button_box.button(QDialogButtonBox.Save).setEnabled(False)
 
-        export_folder = self.get_export_folder_from_dialog()
+        packaged_project_file = Path(self.packagedProjectFileWidget.filePath())
         area_of_interest = (
             self.__project_configuration.area_of_interest
             if self.__project_configuration.area_of_interest
@@ -165,18 +178,21 @@ class PackageDialog(QDialog, DialogUi):
             else QgsProject.instance().crs().authid()
         )
 
-        self.qfield_preferences.set_value("exportDirectoryProject", export_folder)
+        self.qfield_preferences.set_value(
+            "exportDirectoryProject", packaged_project_file.parent
+        )
         self.dirsToCopyWidget.save_settings()
 
         offline_convertor = OfflineConverter(
             self.project,
-            export_folder,
+            packaged_project_file,
             area_of_interest,
             area_of_interest_crs,
             self.qfield_preferences.value("attachmentDirs"),
             self.offliner,
             ExportType.Cable,
             dirs_to_copy=self.dirsToCopyWidget.dirs_to_copy(),
+            export_title=self.packagedProjectTitleLineEdit.text(),
         )
 
         # progress connections
@@ -206,7 +222,7 @@ class PackageDialog(QDialog, DialogUi):
         with a nice link to open the result folder.
         """
         if is_success:
-            export_folder = self.get_export_folder_from_dialog()
+            export_folder = str(Path(self.packagedProjectFileWidget.filePath()).parent)
             result_message = self.tr(
                 "Finished creating the project at {result_folder}. Please copy this folder to "
                 "your QField device."
--- a/qfieldsync/gui/package_dialog.py
+++ b/qfieldsync/gui/package_dialog.py
@@ -136,7 +136,7 @@ class PackageDialog(QDialog, DialogUi):
         else:
             self.show_package_page()
 
-    def get_export_folder(self):
+    def get_export_folder(self) -> str:
         """Get the export folder according to the inputs in the selected"""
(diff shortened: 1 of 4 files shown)


qfieldsync/ui/package_dialog.ui
```diff
@@ -22,15 +22,15 @@
      <item>
       <widget class="QLabel" name="label_2">
        <property name="text">
-        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Project:&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+        <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Exported Project Title:&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
suricactus, line 25: This should not be HTML.

qfieldsync/gui/package_dialog.py
```diff
@@ -189,6 +192,8 @@ def package_project(self):
         try:
             QApplication.setOverrideCursor(Qt.WaitCursor)
             offline_convertor.convert()
+            self.project.setTitle(self.original_title_project)
+            open_project(self.project.fileName())
```
suricactus, line 196: I guess you don't need it in this PR, do you?

qfieldsync/gui/package_dialog.py
```diff
@@ -73,7 +73,8 @@ def __init__(self, iface, project, offline_editing, parent=None):
         self.qfield_preferences = Preferences()
         self.dirsToCopyWidget = DirsToCopyWidget()
         self.__project_configuration = ProjectConfiguration(self.project)
-        self.project_lbl.setText(get_project_title(self.project))
+        self.original_title_project = project.title()
```
suricactus, line 76:
```suggestion
        self.original_project_title = project.title()
```

qfieldsync/gui/package_dialog.py
```diff
@@ -73,7 +73,8 @@ def __init__(self, iface, project, offline_editing, parent=None):
...
+        self.new_project_title.setText(self.original_title_project)
```
suricactus, line 77:
```suggestion
        self.packaged_project_title.setText(self.original_title_project)
```

qfieldsync/gui/package_dialog.py
SeqLaz: Due this was removed on the other [PR](https://github.com/opengisch/qfieldsync/pull/616) now this will be necessary there

qfieldsync/ui/package_dialog.ui
SeqLaz: Good catch!

qfieldsync/ui/package_dialog.ui
```diff
@@ -7,42 +7,61 @@
...
-       </property>
-      </widget>
-     </item>
-    </layout>
+     <property name="text">
+      <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Packaged Project Name&lt;/span&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+     </property>
+     <property name="textFormat">
+      <enum>Qt::RichText</enum>
+     </property>
```
suricactus, line 27:
```suggestion
```

I guess the HTML problem above comes from the fact that the QLabel is `RichText`.

qfieldsync/ui/package_dialog.ui
```diff
@@ -7,42 +7,61 @@
...
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="label_3">
+     <property name="text">
+      <string>&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Packaged Project Title&lt;/span&gt;&lt;/p&gt;</string>
+     </property>
+     <property name="textFormat">
+      <enum>Qt::RichText</enum>
```
suricactus, line 49: Same as above.

qfieldsync/gui/package_dialog.py
```diff
@@ -70,10 +70,15 @@ def __init__(self, iface, project, offline_editing, parent=None):
         self.iface = iface
         self.offliner = QgisCoreOffliner(offline_editing=offline_editing)
         self.project = project
+        self.original_project_path = self.project.fileName()
         self.qfield_preferences = Preferences()
         self.dirsToCopyWidget = DirsToCopyWidget()
         self.__project_configuration = ProjectConfiguration(self.project)
-        self.project_lbl.setText(get_project_title(self.project))
+        self.original_project_title = project.title()
+        self.packaged_project_title.setText(self.original_project_title)
```
suricactus, line 78:
```suggestion
        self.packaged_project_title.setText(project.title())
```

qfieldsync/gui/package_dialog.py
```diff
@@ -70,10 +70,15 @@ def __init__(self, iface, project, offline_editing, parent=None):
         self.iface = iface
         self.offliner = QgisCoreOffliner(offline_editing=offline_editing)
         self.project = project
+        self.original_project_path = self.project.fileName()
```
suricactus, line 73: This is only for internal use, right? Can we make it a "private" property by prefixing with underscore?
```suggestion
        self._original_project_path = self.project.fileName()
```

Also, please move it next to the rest of the changes in this PR a few lines below.

qfieldsync/gui/package_dialog.py
```diff
@@ -70,10 +70,15 @@ def __init__(self, iface, project, offline_editing, parent=None):
...
         self.qfield_preferences = Preferences()
         self.dirsToCopyWidget = DirsToCopyWidget()
         self.__project_configuration = ProjectConfiguration(self.project)
-        self.project_lbl.setText(get_project_title(self.project))
+        self.original_project_title = project.title()
+        self.packaged_project_title.setText(self.original_project_title)
+        self.packaged_project_name.setText(self.project.baseName())
```
suricactus, line 79: What if the project we are trying to package is not saved yet? Imagine you open a new blank project, set all the layers, config etc and then press "package project" without ever saving it to the disk?

qfieldsync/gui/package_dialog.py
```diff
@@ -70,10 +70,15 @@ def __init__(self, iface, project, offline_editing, parent=None):
...
+        self.tmp_project = make_temp_qgis_file(self.project)
```
suricactus, line 80:
```suggestion
        self.tmp_project_filename = make_temp_qgis_file(self.project)
```

qfieldsync/gui/package_dialog.py
```diff
@@ -168,6 +173,14 @@ def package_project(self):
         self.qfield_preferences.set_value("exportDirectoryProject", export_folder)
         self.dirsToCopyWidget.save_settings()
 
+        tmp_path = os.path.dirname(self.tmp_project)
```
suricactus, line 176:
```suggestion
        tmp_project_filepath = Path(self.tmp_project_filename).parent
```

Or whatever is the syntax using `pathlib`. Note that in all the codebase we prefer using `pathlib` over `os.path`.

qfieldsync/gui/package_dialog.py
```diff
@@ -189,6 +202,7 @@ def package_project(self):
         try:
             QApplication.setOverrideCursor(Qt.WaitCursor)
             offline_convertor.convert()
+            open_project(self.original_project_path)
```
suricactus, line 205: I think this is already handled internally in the `OfflineConverter`. It should reopen the original project internally, no? If yes, then we need to make sure the OfflineConverter opens the right project.

suricactus (commented):
Instead of complicating the code in the caller (in this case QFieldSync), we should add this functionality within the OfflineConverter itself. Why? Because the converter already does the copy of the file.

For example here we have hardcoded the name of the output file: https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/offline_converter.py#L276

What we can do is to pass the output name to be a parameter when we construct the `OfflineConverter` class. We can even replace the `export_dir` param with `export_filename` param, and then derive the `export_dir` to be the path to `export_filename`.

suricactus:
Please note that most of the suggestion here need code-wide changes, the suggestions are only pointing to the highlights.

qfieldsync/gui/package_dialog.py
SeqLaz: This is done in this [PR](https://github.com/opengisch/libqfieldsync/pull/94)

SeqLaz:
@suricactus The changes and suggestions were addressed in this [PR](https://github.com/opengisch/libqfieldsync/pull/94)

SeqLaz:
> Please note that most of the suggestion here need code-wide changes, the suggestions are only pointing to the highlights.

@suricactus The changes and suggestions were addressed in this [PR](https://github.com/opengisch/libqfieldsync/pull/94)

suricactus:
Few comments after the libqfieldsync review:
- Drop the "export directory" and export finame in favor of single "export filename" with the full path
- After clicking "Create" OR when changing the value of "export filename", make sure the user is notified there is already a file with that name in that location.
- Prepopulate the "export_filename" with the default value of "export directory" with "/" and current project basename and "_qfield" suffix and ".qgs" extension. You can create a utility function in libqfieldsync `get_export_filename_suggestion(project.fileName())`

qfieldsync/gui/package_dialog.py
```diff
@@ -145,15 +149,43 @@ def get_export_folder_from_dialog(self):
...
+            else:
+                self.show_package_page()
+        else:
+            self.package_project()
+
     def package_project(self):
         self.button_box.button(QDialogButtonBox.Save).setEnabled(False)
 
-        export_folder = self.get_export_folder_from_dialog()
+        export_packaged_project = Path(self.packaged_project_filename.text())
```
suricactus, line 188:
```suggestion
        packaged_project_filename = Path(self.packaged_project_filename.text())
```

qfieldsync/gui/package_dialog.py
```diff
@@ -145,15 +149,43 @@ def get_export_folder_from_dialog(self):
...
         self.stackedWidget.setCurrentWidget(self.packagePage)
 
+    def run_package_project(self):
+        export_packaged_project = Path(self.packaged_project_filename.text())
+
+        if export_packaged_project.exists():
+            reply = QMessageBox.question(
+                self,
+                "File Already Exists",
+                f"'{export_packaged_project.name}' already exists. Do you want to overwrite it?",
```
suricactus, line 174:
```suggestion
                self.tr("File Already Exists"),
                self.tr('File "{}" already exists. Do you want to overwrite it?').format(export_packaged_project.name),
```
don't forget the translation

qfieldsync/ui/package_dialog.ui
```diff
@@ -111,11 +81,33 @@
        <item>
         <widget class="QGroupBox" name="groupBox">
          <property name="title">
-          <string>Export Directory</string>
+          <string>Packaged Project Title</string>
          </property>
          <layout class="QGridLayout" name="gridLayout">
           <item row="0" column="0">
-           <widget class="QLineEdit" name="manualDir"/>
+           <widget class="QLineEdit" name="packaged_project_title"/>
```
suricactus, line 88:
```suggestion
           <widget class="QLineEdit" name="packagedProjectTitleLineEdit"/>
```
Note that the names here are camelCase for historical reasons. Don't mix multiple naming conventions.

qfieldsync/ui/package_dialog.ui
```diff
@@ -111,11 +81,33 @@
...
+           <property name="maximumSize">
+             <size>
+             <width>0</width>
+             <height>0</height>
+             </size>
+           </property>
+         </widget>
+       </item>
+          <item row="0" column="0">
+           <widget class="QLineEdit" name="packaged_project_filename"/>
```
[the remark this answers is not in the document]
suricactus:
```suggestion
           <widget class="QgsFileWidget" name="packagedProjectFileWidget"/>
```

qfieldsync/ui/package_dialog.ui
```diff
@@ -111,11 +81,33 @@
...
+       <item>
+         <widget class="QLineEdit" name="manualDir">
+           <property name="maximumSize">
+             <size>
+             <width>0</width>
+             <height>0</height>
+             </size>
+           </property>
+         </widget>
+       </item>
```
suricactus, line 108: Why do we have the `manualDir` here? Also I think there is some kind of an error with the XML, or I miss somehow the closing `</item>` for line 93. How did you edit this?

suricactus (changes requested):
Not there yet.

SeqLaz:
> Not there yet.

Hey @suricactus please have a look.
I squashed due to the PR had a lot of merge. In the last commit I addressed all the suggestion and some improvements.

qfieldsync/gui/package_dialog.py
```diff
@@ -140,20 +136,51 @@ def setup_gui(self):
...
-        return self.manualDir.text()
+        export_dirname = self.qfield_preferences.value("exportDirectoryProject")
+        if not export_dirname:
+            export_dirname = os.path.join(
+                self.qfield_preferences.value("exportDirectory"),
+                fileparts(QgsProject.instance().fileName())[1],
+            )
+        return QDir.toNativeSeparators(str(export_dirname))
+
+    def get_export_filename_suggestion(self):
```
suricactus, line 149:
```suggestion
    def get_export_filename_suggestion(self) -> str:
```

suricactus, line 149:
```suggestion
    def get_export_filename_suggestion(self) -> str:
```

Please add typing on all new methods you add.

SeqLaz:
Hey @suricactus I added the typing hints on the new functions I added, have a look!

qfieldsync/gui/package_dialog.py
```diff
@@ -140,20 +136,51 @@ def setup_gui(self):
...
+        export_folder = Path(self.get_export_folder())
+        full_project_name_suggestion = export_folder.joinpath(
+            f"{self.project.baseName()}_qfield.qgs"
+        )
+        return str(full_project_name_suggestion)
+
+    def set_export_filename_suggested(self) -> None:
+        self.packagedProjectFileWidget.setFilePath(
+            self.get_export_filename_suggestion()
+        )
```
suricactus, line 159: I failed to find who uses this? In any case, such functions should not exist, it's better to do it explicitly what you do in the function body.

SeqLaz: You are completely right, the function is not needed!

SeqLaz:
Hey @suricactus, I deleted the unnecessary function and joined the functions of export_folder and get_filename_suggestion.

suricactus (approved):
Great work!

suricactus (changes requested):
Ah, just realized, you never bumped libqfieldsync here. Last step before merging. I guess the tests will fail here too.

suricactus:
@SeqLaz you have a conflict now. I will resolve it to unblock the tests.

SeqLaz:
Hey @suricactus I've fixed the tests!.

PR #620, opened 2024-10-23
Use the etag to check whether the file should be synced or not

suricactus:
Drastically speed up projects with a lot of files in them.


nirvn (approved):

nirvn:
@suricactus , we've just gone through a nice etag age of enlightenment  ;)

Issue #640, opened 2025-02-26
Connect to a self-hosted QFieldCloud instance


T4mmi:
Hi, I see that the QFieldCloud URL is [hard-coded](https://github.com/opengisch/qfieldsync/blob/564fed625c38ee643fe7b1dce1e3b0df86cfe049/qfieldsync/core/cloud_api.py#L180) and modifying this string works ... but is there a user-friendly way to setup the plugin to a different/self-hosted instance that I missed ? (like a plugin setting)

[edit] If this requires a feature dev, would a Pull Request be considered or instantly discarded ?

suricactus:
Hi @T4mmi . Pull requests are very welcome, with no promises whether they will be accepted or not. Opening an issue first is a good way to discuss in advance if the feature makes sense to be upstreamed.

As for the very specific issue, we have documented this already with the information you are looking for here: https://docs.qfield.org/get-started/tutorials/advanced-setup-qfc/?h=server#connect-to-a-custom-qfieldcloud-server-in-qfield-and-qfieldsync

T4mmi:
Oh sorry, I missed the double-click part 🤦
> Double-tap on the Nyuki icon (the blue bee QFieldCloud logo).

May I suggest to set it visible by default ?

T4mmi:
Ok ... I'll edit my PR to just add the tooltip then :)

PR #641, opened 2025-02-26
Add tooltip to the QFieldCloudIcon

T4mmi:
see #640


suricactus:
I think it would be easier if you add the tooltip via qtdesigner and make it generic:

"Double click to toggle set QFieldCloud server URL"

Here: https://github.com/opengisch/qfieldsync/blob/master/qfieldsync/ui/cloud_login_dialog.ui#L18

This way there will be 0 python to be written.

T4mmi:
Sorry, I'm really uncomfortable with QtDesigner (not sure I ever used it).  Also since this feature is dynamic (interaction/click changes the UI) it feels like it should be implemented logic-side.

Also, from #640, I'm not sure displaying the target URL should be considered "visual clutter" (we are talking about the "where I'm gonna store my data", and as mentioned it's only visible at first login)

suricactus:
I am closing this PR as inactive.

PR #654, opened 2025-03-25
[WIP] SSO implementation

gounux:
This PR intends to add Single Sign-On login capabilities to QFieldSync.

_Current status is WIP+++, more evolutions to come._

## Requirements

- have a QGIS where [this PR adding extra-tokens to OAuth2 configs](https://github.com/qgis/QGIS/pull/60668) is applied
- have a local QFieldCloud running on [this branch with everything needed for testing SSO with QF/QFS client](https://github.com/opengisch/QFieldCloud/tree/feat/oidc-qfield-testing)

## UI changes

_This is a first draft here, nothing is fixed in stone, very open to discussion and propositions._

- on first login, the dialog only displays the server URL:

image

- when clicking on the `Authenticate` button, the auth methods are fetched from the targeted QFieldCloud server (`/api/v1/auth/providers/` endpoint), and the UI is adapted accordingly :

image

- clicking on `Google` or `OpenID` creates and stores a new `QgsAuthMethodConfig`, that should be used later on when sending some `QNetworkRequest` to the QFieldCloud server, using the `QtNetwork` framework. So we rely on QGIS's `QgsAuthManager` singleton instance to update all requests to QFieldCloud, using the customized OAuth2 config.

More detailed description to come.


qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+        "requestTimeout": 30,
+        "version": 1,
+        "scope": auth_data["scope"],
+    }
+
+    auth_config_str = str(auth_config_dict)
+    auth_config_str = auth_config_str.replace("'", '"')
+    auth_config_str = auth_config_str.replace("False", "false")
+    auth_config_str = auth_config_str.replace("True", "true")
+    auth_config_map = {"oauth2config": auth_config_str}
```
nirvn, line 169: use json.dumps(auth_config_dict) here, it'll give you a proper json string that won't require to to these problematic replaces (e.g. imagine if a token URL has a `False` string in it and the server is case sensitive, we'd break it.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+
+    auth_config = QgsAuthMethodConfig()
+    auth_config.setMethod("OAuth2")
+    auth_config.setVersion(1)
+    auth_config.setName(config_name)
+    auth_config.setUri(related_uri)
+    auth_config.setConfigMap(auth_config_map)
```
nirvn, line 176: I'd avoid using setConfigMap in favor of setConfig.

```suggestion
    auth_config.setConfig("oauth2config", auth_config_str)
```

qfieldsync/core/cloud_api.py
```diff
@@ -467,16 +641,18 @@ def cloud_post(
...
         request.setAttribute(QNetworkRequest.FollowRedirectsAttribute, True)
         request.setHeader(QNetworkRequest.ContentTypeHeader, "application/json")
 
-        if self._token:
-            request.setRawHeader(
-                b"Authorization", "Token {}".format(self._token).encode("utf-8")
-            )
+        if self.auth_method == CloudAuthMethod.SSO:
+            self._add_csrf_cookies_to_request(url, request)
+            self._add_provider_id_header_to_request(request)
```
nirvn, line 653: you'd need this in def cloud_get() too, correct? Recommendation: I'd move this _into_ the def _set_request_auth(request) function, so everything authentication-related happens in a single function. 100% sure it'll prevent mistakes in the future.

qfieldsync/core/cloud_api.py
```diff
@@ -493,16 +669,18 @@ def cloud_put(
...
```
nirvn, line 681: See above, move that into _set_request_auth() so we don't repeat the code.

qfieldsync/core/cloud_api.py
```diff
@@ -519,16 +697,18 @@ def cloud_patch(
...
```
nirvn, line 709: See above, etc. :)

qfieldsync/core/cloud_api.py
```diff
@@ -222,6 +321,21 @@ def set_auth(self, url, **kwargs: str) -> None:
             auth_manager.storeAuthenticationConfig(cfg)
             self.preferences.set_value("qfieldCloudAuthcfg", cfg.id())
 
+    def auth_provider_id(self) -> str:
+        return self.preferences.value("qfieldCloudAuthProviderId")
+
+    def set_auth_provider_id(self, provider_id: str) -> None:
+        self.preferences.set_value("qfieldCloudAuthProviderId", provider_id)
```
nirvn, line 328: In QField, we store that int he QgsAuthMethodConfig itself, which IMHO is a better approach as it ties all the authentication pieces onto one config.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+    auth_config_str = auth_config_str.replace("True", "true")
+    auth_config_map = {"oauth2config": auth_config_str}
+
+    auth_config = QgsAuthMethodConfig()
+    auth_config.setMethod("OAuth2")
+    auth_config.setVersion(1)
+    auth_config.setName(config_name)
+    auth_config.setUri(related_uri)
+    auth_config.setConfigMap(auth_config_map)
+
```
nirvn, line 177: It's also store the provider ID here, i.e. `auth_config.setConfig("qfieldcloud-sso-id", auth_data["id"])`

qfieldsync/core/preferences.py
```diff
@@ -31,6 +32,8 @@ def __init__(self):
         self.add_setting(Dictionary("qfieldCloudLastProjectFiles", Scope.Global, {}))
         self.add_setting(String("qfieldCloudServerUrl", Scope.Global, ""))
         self.add_setting(String("qfieldCloudAuthcfg", Scope.Global, ""))
+        self.add_setting(String("qfieldCloudAuthProviderId", Scope.Global, ""))
```
nirvn, line 35: See comment above about where I think this would be best stored.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -120,15 +144,80 @@ def __init__(
...
+    def on_fetch_auth_request_finished(self) -> None:
+        try:
+            auth_methods = self.network_manager.json_array(self.auth_cap_reply)
+        except CloudException as e:
+            QgsMessageLog.logMessage(
+                f"Could not fetch auth methods from server '{self.serverUrlCmb.currentText()}'",
+                "QFieldSync",
+                Qgis.Warning,
+            )
+            QgsMessageLog.logMessage(str(e), "QFieldSync", Qgis.Critical)
```
nirvn, line 188: I don't think we should log this, there will be plenty of self-hosted QFC servers (and on premise ones) that will not run on the latest QFC version where not having an authentication providers list endpoint is perfectly fine.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -120,15 +144,80 @@ def __init__(
...
+
+            # if google auth available, we need to display the logo for legal reasons.
+            if auth_method["id"] == "google":
+                login_button.setIcon(
+                    QIcon(
+                        os.path.join(
+                            os.path.dirname(__file__), "../resources/google.svg"
+                        )
+                    )
+                )
```
nirvn, line 215: This will need adjusting to align with the QFC PR here (https://github.com/opengisch/QFieldCloud/pull/1192).

It also seems like we will have no choice but to style the button itself (styling colors provided by QFC). You can use the QWidget's setStylesheet on the QPushButton[s] for that.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -195,5 +300,17 @@ def on_login_finished(self) -> None:
...
+    def on_login_with_sso_provider_button_clicked(self, provider_data: dict) -> None:
+        server_url = self.serverUrlCmb.currentText()
+        auth_config = build_oauth2_auth_config_from_cloud_method(
+            provider_data,
+            server_url,
+            "qfieldcloud-sso",
+        )
+        self.network_manager.set_url(server_url)
+        self.network_manager.set_sso_auth_config(auth_config)
+        self.network_manager.set_auth_provider_id(provider_data["id"])
```
nirvn, line 312: See comment on where I think this should go above.

qfieldsync/core/cloud_api.py
gounux: I agree, less duplication as well.

qfieldsync/core/cloud_api.py
gounux: Alright, it would also use less key in settings, which is a good thing IMO.

qfieldsync/gui/cloud_login_dialog.py
gounux: Do you mean instead of logging, we should just enable the good ol' username/password ?

nirvn: @gounux , yeah exactly, I don't think we should treat the absence of that API endpoint as something abnormal that's worthy of warning users about. If it's not there, we just enable username and password credentials.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+    Returns:
+        QgsAuthMethodConfig: QGIS auth config for the provided method.
+    """
+    auth_config_dict = {
+        "method": "oauth2",
+        "accessMethod": 0,
+        "grantFlow": auth_data["grant_flow"],
+        "configType": 1,
+        "clientId": auth_data["client_id"],
+        "clientSecret": auth_data["client_secret"],
```
lukasgraf, line 150: So, this is something that has changed from my PR opengisch/QFieldCloud#1192 to the implementation that is on the `feat/oidc-qfield-testing` branch - the implementation in the PR doesn't send any `client_secret` any more.

This is because this endpoint is public, and we cannot publicly disclose the client secret. Which in turn means, we actually don't want to use client secrets at all. Which is no problem, because the alternative, PKCE, works just as well, and with PKCE a client secret isn't needed at all.

So essentially line 150 here can and should be removed.

However: The Keycloak setup on the `feat/oidc-qfield-testing` branch is already set up to use PKCE. However, the Google OAuth2 app that Matthias set up is currently still configured to use a client secret. Meaning that it will break once we don't use the secret any more. I would suggest that I ask him on Monday to switch the Google app to PKCE.

qfieldsync/core/cloud_api.py
```diff
@@ -133,9 +201,21 @@ def __init__(self, parent=None) -> None:
...
         url = self.preferences.value("qfieldCloudServerUrl")
+
         # we should always use the QgsNetworkAccessManager instance, otherwise ssl handling is impossible
         self._nam = QgsNetworkAccessManager.instance()
 
+        self.auth_method = CloudAuthMethod(
+            self.preferences.value("qfieldCloudAuthMethod")
+        )
+        pref_auth_config_id = self.preferences.value("qfieldCloudAuthcfg")
+        if pref_auth_config_id and pref_auth_config_id != "":
```
lukasgraf, line 212: This can be simplified to just 
```python
if pref_auth_config_id:
```

`if value` in Python will evaluate to `False` for `None`, `False`, `0`, `""`, `[]`, `{}`, ... and to `True` otherwise. It's usually considered more idiomatic Python, except in cases where you actually need to distinguish `None` from the empty string, for example.

qfieldsync/core/cloud_api.py
lukasgraf: I like that suggestion a lot. The `provider_id` is inherently tied to the auth config, that's a good place for it to live 👍

qfieldsync/core/cloud_api.py
```diff
@@ -244,22 +358,33 @@ def server_url(self):
...
+            cfg = self.auth()
+
+            server_url = cfg.uri() or self.url
+            username = cfg.config("username")
+            password = cfg.config("password")
 
-        server_url = cfg.uri() or self.url
-        username = cfg.config("username")
-        password = cfg.config("password")
+            if username and password:
```
lukasgraf, line 374: This works, but "only by accident":
For `username = "foo"` and `password = "bar"`, the expression `username and password` will evaluate to `"bar"`.

So you're essentially doing `if "bar"` in that statement. Which works correctly, but is probably not what you intended.

A more idiomatic way in Python would be:
```python
if all([username, password]):
```
using the [`all()` builtin](https://docs.python.org/3/library/functions.html#all). That evaluates all the elements in the sequence to boolean, and returns `True` if all of them evaluate to `True`. If not, it will short-circuit and return `False` as soon as it encounters the first falsy value.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+
+def build_oauth2_auth_config_from_cloud_method(
+    auth_data: dict, related_uri: str, config_name: str = "qfieldcloud-sso"
+) -> QgsAuthMethodConfig:
+    """Builds a QgsAuthMethodConfig from a method provided by QFieldCloud's auth capabilities.
+
+    Args:
+        auth_data (dict): dict describing an auth method.
+        related_uri (str): URI of the generated auth config.
+        config_name (str, optional): name of the generated auth config. Defaults to "qfieldcloud-oauth2".
```
lukasgraf, line 139: -> `Defaults to "qfieldcloud-sso"`

lukasgraf (commented):
Nice! 🎉 A few minor comments in addition to what @nirvn already wrote, but otherwise LGTM! 👍

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -198,21 +201,55 @@ def on_fetch_auth_methods_finished(self) -> None:
...
+        pixmap = QPixmap(32, 32)
+        pixmap.fill(Qt.GlobalColor.transparent)
+        painter = QPainter(pixmap)
+        renderer.render(painter)
+        painter.end()
+
+        button.setIcon(QIcon(pixmap))
+
+    def get_qgis_theme(self, available_themes: list[str]) -> str:
+        # TODO: get light/dark theme depending on current QGIS theme
```
gounux, line 250: Would you know a trick to get the `dark` or `light` theme based on current QGIS settings / theme ?

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -198,21 +201,55 @@ def on_fetch_auth_methods_finished(self) -> None:
...
+            f"background-color: {theme.get('color_fill')}; border-color: {theme.get('color_stroke')}; color: {theme.get('color_text')};"
+        )
+
+        # download svg logo and apply it to button
+        icon_url = theme.get("logo")
+        icon_req = QNetworkRequest(QUrl(icon_url))
+        icon_reply = QgsNetworkAccessManager.instance().get(icon_req)
+        icon_reply.finished.connect(
+            partial(self.on_get_svg_logo_reply_finished, icon_reply, button)
+        )
```
gounux, line 232: Am wondering if this (`gui` folder) is the best place for such network considerations, :thinking: ?

gounux (commented):
PR description updated with screenshots: SSO providers buttons are dynamically generated and styled.

Review with a few questions I'm asking myself ATM, maybe you can help @nirvn @lukasgraf ?

In the meantime, the work continues to make sure there is no regression in the other parts / features of the plugin, regarding QFC/QFS workflow.

qfieldsync/core/cloud_api.py
```diff
@@ -270,19 +395,52 @@ def login(self, username: str, password: str) -> Optional[QNetworkReply]:
...
+            authcfg = self.preferences.value("qfieldCloudAuthcfg")
+            auth_manager = QgsApplication.authManager()
+            auth_manager.clearCachedConfig(authcfg)
+            auth_manager.removeAuthenticationConfig(authcfg)
+            self.auth_method = CloudAuthMethod.NONE
+            self.preferences.set_value("qfieldCloudAuthcfg", "")
+            self.auth_config = None
+            self.preferences.set_value("qfieldCloudAuthMethod", self.auth_method.value)
+            self.logout_success.emit()
+            return None
```
suricactus, line 443: If I understand the logic here properly, you never invalidate the token? I think this is a problem, I would expect that logout invalidates tokens and nobody can use them anymore?

qfieldsync/core/cloud_api.py
```diff
@@ -270,19 +395,52 @@ def login(self, username: str, password: str) -> Optional[QNetworkReply]:
                 "password": password,
             },
         )
-        reply.finished.connect(lambda: self._on_login_finished(reply))
+        reply.finished.connect(lambda: self._on_login_with_credentials_finished(reply))
+
+        return reply
+
+    def _get_cloud_user_info(self) -> Optional[QNetworkReply]:
```
suricactus, line 402: Would you mind moving the private methods somewhere in the bottom or at least not splitting the two important methods `login_with_...` in two.

qfieldsync/core/cloud_api.py
nirvn: Also, lesson learned from the QField side of things: when logging out from an SSO session, we need to clear network cookies, otherwise we can't hop back into user/password credentials.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,69 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
     return CloudException(reply, Exception(message))
 
 
+class CloudAuthMethod(Enum):
+    NONE = 0
+    CREDENTIALS = 1
+    SSO = 2
+
+
+def build_oauth2_auth_config_from_cloud_method(
```
suricactus, line 131: I do not follow `from_cloud_method` suffix here? Is it needed, or can we rename to something that is easier to follow?

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,69 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+    auth_data: dict, related_uri: str, config_name: str = "qfieldcloud-sso"
```
suricactus, line 132: Is the `qfieldcloud-sso` a string under our control? If yes, the whole QField* codebase has a strong preference on `_` over `-`.

qfieldsync/core/cloud_api.py
```diff
@@ -182,19 +260,36 @@ def server_urls() -> List[str]:
...
         ]
 
+    def username(self) -> Optional[str]:
+        if self.auth_method == CloudAuthMethod.CREDENTIALS:
+            return self.auth().config("username")
+        elif self.auth_method == CloudAuthMethod.SSO:
+            if self.current_username:
+                return self.current_username
+            else:
+                return None
```
suricactus, line 270: Isn't it the same if you just write `self.current_username`? I know it might be an empty string, but isn't it a worse error?

qfieldsync/gui/cloud_login_dialog.py
suricactus: Why don't you use the `cloud_api.py`?

qfieldsync/core/cloud_api.py
```diff
@@ -182,19 +260,36 @@ def server_urls() -> List[str]:
             "https://localhost:8002/",
         ]
 
+    def username(self) -> Optional[str]:
```
suricactus, line 263: all the methods we have in the codebase are having some kind of a verb, so `get_username` is a good candidate. Another option would be to convert it to dynamic `@property`. But don't mix the two.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -45,6 +57,11 @@
 class CloudLoginDialog(QDialog, CloudLoginDialogUi):
     instance = None
 
+    _fetch_auth_methods_timer: QTimer = QTimer()
+
+    _credentials_auth_method: Optional[dict] = None
+    _sso_auth_methods: Optional[dict] = {}
```
suricactus, line 63: 1) can we avoid `Optional` whereever possible? In this case if we just have an empty dict, it will be perfectly fine.
2) can we have a bit of typing for the dict? Ideally a `TypedDict`, minimum should be `dict[str, Any]` or whatever it is.

suricactus: Actually I find it a bit strange that we store the state of the currently available auth methods in th gui.

I guess we need to add a new class where all this login lives? Or if it is not clear what should be the shape of this class, keep everything in cloud_api?

qfieldsync/gui/cloud_login_dialog.py
gounux: Alright, thanks! Let me try that

qfieldsync/core/cloud_api.py
gounux: > If I understand the logic here properly, you never invalidate the token? I think this is a problem, I would expect that logout invalidates tokens and nobody can use them anymore?

Good point, indeed, let's also keep in mind that the `QgsAuthMethodConfig` is deleted from the `QgsApplication.authManager()`.

If you know a way of invalidating the token, I'd be interested !

suricactus: Check the logout method in Django. In the end we get the django session cookies, no? Or it is something that allauth is doing?  Check with the other sso gurus you have :)

qfieldsync/resources/google.svg
suricactus, line 1: I imagined that we can use the logo image as served from the QFieldCloud api?

lukasgraf: Exactly, that was the idea - the `auth/providers` endpoint will return a publicly accessible asset URL to the logo for the provider. e.g. `logo_url = provider["styles"]["light"]["logo"]` for a logo that works on a light background.

gounux: That is absolutely true, this icon is not used anymore. Let me delete it.

qfieldsync/core/cloud_api.py
nirvn: Which, incidentally, is what we ended up doing in QField: https://github.com/opengisch/QField/blob/master/src/core/qfieldcloudconnection.cpp#L366-L383 :)

qfieldsync/gui/cloud_login_dialog.py
lukasgraf: I agree with this, but then again QT development is not my strong suit 😉 

But as a general remark:
I think the clients should persist as little as possible about the authentication methods/provider, and not do it for longer than technically necessary. Technically speaking, the clients should expect that a chosen authentication method and its parameters stays valid for the duration of the users session, but not any longer. Auth methods might and will change, and so will their parameters.

So in simplified terms, I think the clients should:
- Present the user with the options for login based on the response from the `auth/providers` endpoint
- If the user selects a provider of type `oauth2`, create the appropriate auth config based on the info from the response for that provider, and set it as the active auth config (but not store any info about the other providers that weren't used).
- On logout, discard the auth config (and cookies) completely. Create it again on next login.

qfieldsync/core/cloud_api.py
gounux: > So, to summarize, what I think clients should do on logout:
> 
>     * Clear cookies
> 
>     * Purge auth config

I believe this is what is being done so far, thanks !

qfieldsync/gui/cloud_login_dialog.py
gounux: When you click on `Login with ...`, a `QgsAuthMethodConfig` is created and stored, based on the JSON provided by QFieldCloud (`/auth/providers` endpoint). This config is removed from manager and QGIS auth methods storage when logging out.

> So in simplified terms, I think the clients should:
> 
>     * Present the user with the options for login based on the response from the `auth/providers` endpoint
> 
>     * If the user selects a provider of type `oauth2`, create the appropriate auth config based on the info from the response for that provider, and set it as the active auth config (but not store any info about the other providers that weren't used).
> 
>     * On logout, discard the auth config (and cookies) completely. Create it again on next login.

I think the code, so far, is aligned with that!

qfieldsync/core/cloud_api.py
```diff
@@ -270,19 +394,54 @@ def login(self, username: str, password: str) -> Optional[QNetworkReply]:
...
+        This is typically called as a first request
+
+        Returns:
+            Optional[QNetworkReply]: QNetworkReply from the QFieldCloud server.
+        """
+        # don't login multiple times
+        if self.is_login_active:
+            return
+
+        self.is_login_active = True
```
gounux, line 415: This mechanism has been mimicked from pre-existing login with credentials: it is convenient and useful for freezing the UI during login. _But_, when logging in with SSO providers, we are not sure the user would perform the few clicks required in the browser, so if the user stops the SSO login process, we might have a frozen UI.

We could consider removing this login locking process, but removing it could create a mess if clicking for instance on several SSO providers.

qfieldsync/core/cloud_api.py
```diff
@@ -129,14 +129,18 @@ class CloudAuthMethod(Enum):
 
 
 def build_oauth2_auth_config(
-    auth_data: dict, related_uri: str, config_name: str = "qfieldcloud_sso"
+    auth_data: dict,
```
suricactus, line 132: Do we know more about this dict?

qfieldsync/core/cloud_api.py
```diff
@@ -129,14 +129,18 @@ class CloudAuthMethod(Enum):
...
+    related_uri: str,
+    config_name: str = "qfieldcloud_sso",
+    persist: bool = True,
```
suricactus, line 135: Do we have control over this parameter name?  Can we prefix with SHOULD or other boolean indicator?

qfieldsync/core/cloud_api.py
gounux: It's the raw one coming from QFC, on the `/auth/providers` endpoint.

gounux: I'm not sure we can go further than `dict[str, Any]`

nirvn:
@gounux , tested the plugin just now and looked at the code, working like a charm, thanks. 

I'd say this feels ready to merge (after we push a 4.13.2 release :) ).

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -195,5 +350,17 @@ def on_login_finished(self) -> None:
...
+    def on_login_with_sso_provider_button_clicked(self, provider_data: dict) -> None:
+        server_url = self.serverUrlCmb.currentText()
+        auth_config = build_oauth2_auth_config(
+            provider_data,
+            server_url,
+            should_persist=self.rememberMeCheckBox.isChecked(),
+        )
+        self.network_manager.set_url(server_url)
+        self.network_manager.set_auth_method(CloudAuthMethod.SSO)
+        self.network_manager.set_sso_auth_config(auth_config)
```
gounux, line 362: Here we store the `QgsAuthMethodConfig` before being sure the user is authenticated (will come a bit later with a GET request to the `auth/user/`, thus triggering the QGIS OAuth2 login mechanism in the browser.

gounux (commented):
> @gounux , tested the plugin just now and looked at the code, working like a charm, thanks.

Thanks @nirvn !

I guess there are still some details to polish, for instance if you click `Login with XYZ`, then the browser pops up, and if you don't complete the login in the browser, you might have a `QgsAuthConfigMethod` stored and set for the `QgsApplication.authManager()` to be used, but not authenticated yet. So it is possible that, in the current state of the code, the user get stucked with, even when spamming the `Login with XYZ` button:

image

Basically, clicking on a `Login with...` button create a config in QGIS authentication settings, even if the login process doesn't go all the way. So there might be sevral auth config, which is also something that can be made better I believe.

qfieldsync/gui/cloud_login_dialog.py
gounux: But maybe this can be done in a follow-up PR.

gounux:
@nirvn @suricactus should I squash the commits now ?

qfieldsync/ui/cloud_login_dialog.ui
```diff
@@ -22,9 +22,6 @@
        <verstretch>0</verstretch>
       </sizepolicy>
      </property>
-     <property name="toolTip">
-      <string>Double-click me to show or hide advanced options</string>
-     </property>
```
suricactus, line 27: I guess we lost this by accident?

gounux: Indeed, Qt Designer driving too fast..

Fixed, thanks for spotting.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -151,14 +291,28 @@ def authenticate(self) -> None:
...
+            self.network_manager.login_with_sso()
 
-        if not cfg.config("token") or not self.parent():
+        elif not cfg.config("token") or not self.parent():
             self.show()
             self.raise_()
             self.activateWindow()
 
-    def on_login_button_clicked(self) -> None:
+    def on_server_url_edit_text_changed(self) -> None:
```
suricactus, line 307: You better describe what we are doing here as a comment.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -120,15 +146,127 @@ def __init__(
...
+            self.set_sso_provider_button_style(auth_method.get("styles"), login_button)
+
+            login_button.clicked.connect(
+                partial(self.on_login_with_sso_provider_button_clicked, auth_method)
+            )
+            self.signInUsernameGroupBox.layout().addWidget(login_button)
+            self._sso_login_buttons.append(login_button)
+
+    def set_sso_provider_button_style(
+        self, style_data: dict, button: QPushButton
```
suricactus, line 217: a bit more about the `dict`?

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+class CloudAuthMethod(Enum):
+    NONE = 0
+    CREDENTIALS = 1
+    SSO = 2
+
+
+def build_oauth2_auth_config(
+    auth_data: dict[str, Any],
+    related_uri: str,
+    config_name: str = "qfieldcloud_sso",
```
suricactus, line 134: Is this free form text, or it has a subset of values? If possible, we can turn it into a `Literal`, if not, it's ok.

qfieldsync/core/cloud_api.py
```diff
@@ -114,13 +122,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
+    should_persist: bool = True,
```
suricactus, line 135:
```suggestion
    should_persist_token: bool = True,
```
or
```suggestion
    is_token_persisted: bool = True,
```
I think it is better?

qfieldsync/core/cloud_api.py
```diff
@@ -133,9 +201,21 @@ def __init__(self, parent=None) -> None:
...
         # we should always use the QgsNetworkAccessManager instance, otherwise ssl handling is impossible
         self._nam = QgsNetworkAccessManager.instance()
 
+        self.auth_method = CloudAuthMethod(
+            self.preferences.value("qfieldCloudAuthMethod")
+        )
+        pref_auth_config_id = self.preferences.value("qfieldCloudAuthcfg")
+        if pref_auth_config_id:
+            auth_config = QgsAuthMethodConfig()
+            _, config = QgsApplication.authManager().loadAuthenticationConfig(
```
suricactus, line 214:
```suggestion
            _whatever, config = QgsApplication.authManager().loadAuthenticationConfig(
```

I find it nice when one names unused variables, so the reader does need to check the API docs. In this case `whatever` should be changed with whatever the value of the variable represents.

qfieldsync/core/cloud_api.py
```diff
@@ -182,19 +264,33 @@ def server_urls() -> List[str]:
...
+
     def auth(self) -> QgsAuthMethodConfig:
         auth_manager = QgsApplication.authManager()
 
         if not auth_manager.masterPasswordHashInDatabase():
             return QgsAuthMethodConfig()
 
         authcfg = self.preferences.value("qfieldCloudAuthcfg")
+
+        if not authcfg or authcfg == "":
```
suricactus, line 282:
```suggestion
        if not authcfg:
```
Aren't they almost the same? Yes, now you also work with support authconfig to be `0` (int), but I don't think we should support that either.

qfieldsync/core/cloud_api.py
```diff
@@ -182,19 +264,33 @@ def server_urls() -> List[str]:
...
+            return QgsAuthMethodConfig()
+
         cfg = QgsAuthMethodConfig()
         auth_manager.loadAuthenticationConfig(authcfg, cfg, True)
 
         return cfg
 
     def set_auth(self, url, **kwargs: str) -> None:
+        if self.auth_method != CloudAuthMethod.CREDENTIALS:
```
suricactus, line 291: I would change that to `assert`, as far as I see the caller already checks if it is the right auth method.

qfieldsync/core/cloud_api.py
```diff
@@ -222,6 +318,27 @@ def set_auth(self, url, **kwargs: str) -> None:
...
             self.preferences.set_value("qfieldCloudAuthcfg", cfg.id())
 
+    def auth_provider_id(self) -> str:
+        """
+        Returns the provider ID, required for sso auth to QFieldCloud.
+        Should be stored in the QgsAuthMethodConfig with the "qfieldcloud_sso_id" key.
+        """
+        auth_method = self.auth()
+        if not auth_method:
+            return ""
```
suricactus, line 328:
```suggestion

        if not auth_method:
            return ""

```

A bit of air. Make sure the early returns are prominently surrounded by empty lines.

qfieldsync/core/cloud_api.py
```diff
@@ -364,7 +531,7 @@ def delete_file(self, filename: str) -> QNetworkReply:
 
     def set_token(self, token: str, update_auth: bool = False) -> None:
         """Sets QFieldCloud authentication token to be used by all the following requests. Set to empty string to disable token authentication."""
-        if update_auth:
+        if update_auth and self.auth_method == CloudAuthMethod.CREDENTIALS:
```
suricactus, line 534: Shall we add `else` that throws if the auth_method is not credentials?

qfieldsync/core/cloud_api.py
```diff
@@ -374,9 +541,35 @@ def set_token(self, token: str, update_auth: bool = False) -> None:
...
         self.token_changed.emit()
 
+    def is_authenticated(self) -> bool:
+        if self.auth_method == CloudAuthMethod.CREDENTIALS:
+            return self.has_token()
+        if self.auth_method == CloudAuthMethod.SSO:
+            # If there is no username, it means a request to get it
+            # must be done before being considered authenticated.
+            return self.auth_config and self.current_username
+        return False
```
suricactus, line 551:
```suggestion
        if self.auth_method == CloudAuthMethod.NONE:
            return False
        elif self.auth_method == CloudAuthMethod.CREDENTIALS:
            return self.has_token()
        elif self.auth_method == CloudAuthMethod.SSO:
            # If there is no username, it means a request to get it
            # must be done before being considered authenticated.
            return self.auth_config and self.current_username
        else:
            raise NotImplementedError("Unknown authentication method {self.auth_method}!")
```

I think this is more future proof.

qfieldsync/core/cloud_api.py
```diff
@@ -680,6 +864,32 @@ def _on_login_finished(self, reply: QNetworkReply) -> None:
...
+        if payload["avatar_url"]:
+            suffix = payload["avatar_url"].rsplit(".")[-1]
+            avatar_filename = tempfile.mktemp(suffix=f".{suffix}")
+            reply = self.get_file(
+                QUrl(payload["avatar_url"]),
+                avatar_filename,
+            )
+            reply.finished.connect(
+                lambda: self._on_avatar_download_finished(reply, avatar_filename)
+            )
```
suricactus, line 889:
```suggestion

        if payload["avatar_url"]:
            suffix = payload["avatar_url"].rsplit(".")[-1]
            avatar_filename = tempfile.mktemp(suffix=f".{suffix}")
            reply = self.get_file(
                QUrl(payload["avatar_url"]),
                avatar_filename,
            )
            reply.finished.connect(
                lambda: self._on_avatar_download_finished(reply, avatar_filename)
            )

```

A little bit of air :)

qfieldsync/core/cloud_api.py
```diff
@@ -733,6 +943,29 @@ def _clear_cloud_cookies(self, url: QUrl) -> None:
...
+        """
+        Adds CSRF cookie to a network request.
+        Required for some QFieldCloud requests, like POST/PUT/PATCH/DELETE.
+        """
+        cookie_jar = self._nam.cookieJar()
+        cookies = cookie_jar.cookiesForUrl(url)
+
+        csrftoken_cookie = next(
+            (cookie for cookie in cookies if cookie.name() == b"csrftoken"), None
+        )
```
suricactus, line 956: Please rewrite imperatively.

qfieldsync/core/cloud_api.py
```diff
@@ -733,6 +943,29 @@ def _clear_cloud_cookies(self, url: QUrl) -> None:
...
+        if csrftoken_cookie:
+            request.setRawHeader(b"X-CSRFToken", csrftoken_cookie.value())
+            request.setRawHeader(b"Referer", url.toEncoded())
+
+    def _add_provider_id_header_to_request(self, request: QNetworkRequest) -> None:
+        """Adds the current provider id to a request, if any. Sometimes required by QFC middleware."""
+        provider_id = self.auth_provider_id()
+
+        if provider_id and len(provider_id) > 0:
+            request.setRawHeader(b"X-QFC-IDP-ID", QByteArray(provider_id.encode()))
```
suricactus, line 967: Do not hardcode the header. Best case should come from QFC, worst case should be a const.

qfieldsync/core/cloud_api.py
```diff
@@ -733,6 +943,29 @@ def _clear_cloud_cookies(self, url: QUrl) -> None:
...
+        """
+        cookie_jar = self._nam.cookieJar()
+        cookies = cookie_jar.cookiesForUrl(url)
+
+        csrftoken_cookie = next(
+            (cookie for cookie in cookies if cookie.name() == b"csrftoken"), None
+        )
+
+        if csrftoken_cookie:
+            request.setRawHeader(b"X-CSRFToken", csrftoken_cookie.value())
```
suricactus, line 959: Do not hardcode the header. Best case should come from QFC, worst case should be a const.

suricactus (changes requested):
Code looks good, great job! A few comments and one general suggestion: please write all new private methods with `_` in front, so we know what is the public API of a class .

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -82,25 +99,34 @@ def __init__(
         self.preferences = Preferences()
         self.network_manager = network_manager
 
-        self.buttonBox.button(QDialogButtonBox.Ok).setText(self.tr("Sign In"))
-        self.buttonBox.button(QDialogButtonBox.Ok).clicked.connect(
-            self.on_login_button_clicked
+        self.buttonBox.setEnabled(False)
+        self.buttonBox.hide()
+
+        self._fetch_auth_methods_timer.setInterval(750)
```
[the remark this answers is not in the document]
gounux: This is the same behaviour as in QField, actually we use a timer not to spam the server with request on each textchanged event. ping @nirvn

qfieldsync/core/cloud_api.py
gounux: I personally prefer the `should` since we just pass it to the QGIS auth manager.

qfieldsync/core/cloud_api.py
```diff
@@ -733,6 +955,31 @@ def _clear_cloud_cookies(self, url: QUrl) -> None:
...
+        """
+        Adds CSRF cookie to a network request.
+        Required for some QFieldCloud requests, like POST/PUT/PATCH/DELETE.
+        """
+        cookie_jar = self._nam.cookieJar()
+        cookies = cookie_jar.cookiesForUrl(url)
+
+        csrftoken_cookie = None
+        for cookie in cookies:
+            if cookie.name() == b"csrftoken":
```
suricactus, line 968: I guess we want to move the cookies name to a constant too?

qfieldsync/core/cloud_api.py
```diff
@@ -270,22 +403,61 @@ def login(self, username: str, password: str) -> Optional[QNetworkReply]:
...
+            self.auth_config = None
+            self.current_username = None
+            self.preferences.set_value("qfieldCloudAuthMethod", self.auth_method.value)
+            self._clear_cloud_cookies(QUrl(self.url))
+            self.logout_success.emit()
+            return None
 
         return reply
 
+    def get_basic_reply(self, resource_url: str) -> QNetworkReply:
```
suricactus, line 457: I am not sure why we need this new method? I guess a docstring can explain it?

qfieldsync/core/cloud_api.py
```diff
@@ -48,6 +56,10 @@
 from qfieldsync.core.preferences import Preferences
 from qfieldsync.utils.qt_utils import strip_html
 
+HTTP_HEADER_CSRF_TOKEN = b"X-CSRFToken"
+HTTP_HEADER_REFERER = b"Referer"
+HTTP_HEADER_IDP_ID = b"X-QFC-IDP-ID"
```
suricactus, line 61: What is the purpose of this header?

gounux: I guess the idea is to know which provider is being used. It is required by the QGIS auth middleware in QFC: https://github.com/opengisch/QFieldCloud/blob/4571b06c3d2f8cdb0a2cd908ab48922ed6b0fe16/docker-app/qfieldcloud/core/middleware/qgis_auth.py#L76

qfieldsync/core/cloud_api.py
```diff
@@ -182,19 +268,32 @@ def server_urls() -> List[str]:
             "https://localhost:8002/",
         ]
 
+    def get_username(self) -> Optional[str]:
+        if self.auth_method == CloudAuthMethod.CREDENTIALS:
+            return self.auth().config("username")
+        elif self.auth_method == CloudAuthMethod.SSO:
+            return self.current_username
+        return None
```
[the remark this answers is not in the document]
gounux: Thanks!

qfieldsync/core/cloud_api.py
suricactus: ok

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -120,15 +148,118 @@ def __init__(
...
+        button.setStyleSheet(
+            """
+            background-color: {color_fill};
+            border-color: {color_stroke};
+            color: {color_text};
+            """.format(
+                # NOTE Ideally we want to escape the values before we inject them into CSS
+                color_fill=theme.get("color_fill"),
+                color_stroke=theme.get("color_stroke"),
+                color_text=theme.get("color_text"),
```
suricactus, line 238: What happens if the keys do not exist and the value is `None`? Shall we add something like `auto` or other default?

qfieldsync/gui/utils.py
```diff
@@ -35,3 +39,22 @@ def set_available_actions(combobox, actions, default_action):
 
         if action == default_action:
             combobox.setCurrentIndex(combobox.count() - 1)
+
+
+def extract_theme_from_qgis_settings() -> str:
```
suricactus, line 44:
```suggestion
def extract_theme_from_qgis_settings() -> Literal["light", "dark"]:
```

qfieldsync/gui/utils.py
```diff
@@ -35,3 +39,22 @@ def set_available_actions(combobox, actions, default_action):
...
+        "light" or "dark", based on user's current QGIS settings.
+    """
+    qgis_theme = QgsApplication.instance().themeName()
+    if qgis_theme == "Night Mapping":
+        return "dark"
+    if qgis_theme == "Blend of Gray":
+        return "light"
+    color = QWidget().palette().color(QPalette.Window)
+    if (color.red() + color.green() + color.blue()) / 3 < 120:
+        return "dark"
```
suricactus, line 59:
```suggestion

    if qgis_theme == "Night Mapping":
        return "dark"

    if qgis_theme == "Blend of Gray":
        return "light"

    color = QWidget().palette().color(QPalette.Window)

    if (color.red() + color.green() + color.blue()) / 3 < 120:
        return "dark"

```
air :)

qfieldsync/gui/cloud_login_dialog.py
gounux: That is right, we can add default values.

duke-nyuki:
Task linked: [QF-5434 Allow login via SSO from QFieldSync](https://app.clickup.com/t/2192114/QF-5434)

qfieldsync/core/cloud_api.py
gounux: Added a check at the start of the method.

suricactus:
Shall we consider merging this at this stage?

gounux:
> Shall we consider merging this at this stage?

I guess so :

- polishing can be made in a follow-up PR.
- current state makes it retro-compatible with older QFC instances.

Apologies for the late reply BTW

suricactus:
Thanks! Then please resolve the conflict and we are ready totest.

nirvn:
Exciting!

gounux:
> Thanks! Then please resolve the conflict and we are ready totest.

Done ! Also rebased, I guess ~45 commits is my personal record :D

qfieldsync/core/cloud_api.py
```diff
@@ -122,13 +130,73 @@ def from_reply(reply: QNetworkReply) -> Optional[CloudException]:
...
 
 
+class CloudAuthMethod(Enum):
+    NONE = 0
+    CREDENTIALS = 1
+    SSO = 2
+
+
+def build_oauth2_auth_config(
+    auth_data: dict[str, Any],
```
suricactus, line 140:
```suggestion
    auth_data: Dict[str, Any],
```
Unfortunately we support some more ancient Python versions and we should use `Dict` instead of `dict` in QFieldSync :/

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -195,5 +350,17 @@ def on_login_finished(self) -> None:
         self.rememberMeCheckBox.setEnabled(False)
         self.done(QDialog.Accepted)
 
+    def on_login_with_sso_provider_button_clicked(self, provider_data: dict) -> None:
```
suricactus, line 353:
```suggestion
    def on_login_with_sso_provider_button_clicked(self, provider_data: Dict[..., ...]) -> None:
```
Missing key and value types for the dict. I guess `Dict[str, Any]`?

qfieldsync/core/cloud_api.py
```diff
@@ -252,22 +373,33 @@ def server_url(self):
...
 
-    def login(self, username: str, password: str) -> Optional[QNetworkReply]:
-        """Login to QFieldCloud"""
+        elif self.auth_method == CloudAuthMethod.SSO:
+            self.login_with_sso()
+
+    def login_with_credentials(
+        self, username: str, password: str
+    ) -> Optional[QNetworkReply]:
+        """Login to QFieldCloud with classic login/password method"""
```
suricactus, line 399:
```suggestion
        """Login to QFieldCloud with login/password method"""
```
Just to prevent someone to confuse it with QFieldCloud classic.

qfieldsync/core/cloud_api.py
suricactus: Sorry I had to reopen this, still not clear to me?

qfieldsync/core/cloud_api.py
```diff
@@ -702,7 +934,7 @@ def get_last_login_error(self) -> str:
         suggest_forgotten_password = True
         error_str = ""
 
-        if self._login_error:
+        if hasattr(self, "_login_error"):
```
suricactus, line 937: I find using `if self._login_error` (probably with a check if it is `None`) a bit cleaner than checking with `hasattr`, is there a specific reason why we cannot guarantee that `_login_error` is present as a property?

gounux: I faced this `self._login_error` not being an attribute in some cases. Maybe initializing in the constructor would help ?

qfieldsync/core/cloud_api.py
gounux: This will be the name of the `QgsAuthMethodConfig` stored in the manager. So "free form text", even if the place where this function is called does not specify it.

qfieldsync/core/cloud_api.py
```diff
@@ -278,22 +412,66 @@ def login(self, username: str, password: str) -> Optional[QNetworkReply]:
...
-        return self.cloud_get("auth/user/", {"token": token})
+    def login_with_sso(self) -> Optional[QNetworkReply]:
+        """
+        Login to QFieldCloud with social provider (e.g. Google).
+        This basically sends a GET user info request to QFC.
+        Which means the OAuth2 QgsAuthMethodConfig will trigger auth in the browser, if required.
+        """
+        return self._get_cloud_user_info()
+
+    def _get_cloud_user_info(self) -> Optional[QNetworkReply]:
```
suricactus, line 427: Why here we return `Optional`, since `cloud_get` does not return optional?

qfieldsync/core/cloud_api.py
```diff
@@ -252,22 +375,33 @@ def server_url(self):
 
         return re.sub(r"([^:]/)(/)+", r"\1", url)
 
+    def get_auth_capabilities(self) -> QNetworkReply:
+        """Get authentication capabilities from a server"""
+        reply = self.cloud_get("auth/providers/")
+        return reply
+
     def auto_login_attempt(self):
```
suricactus, line 383:
```suggestion
    def auto_login_attempt(self) -> None:
```
Feel free to ignore if you don't have other work on this PR

gounux:
> I want to warn you that once #720 is merged you might have a bit of conflicts and required code changes. Apologies for the extra work on your side, I hope it's for the best of the project.

Alright, noted thanks for warning. I can't complain regarding some ruff config ;) Once it's merged I can do the rebase.

suricactus:
@gounux I am happy to merge with or without the change that @lukasgraf suggests above, but make sure you don't forget it.

lukasgraf (approved):

m-kuhn:
🎉

PR #657, opened 2025-04-09
[QFieldSync] When logging to a QFC account add the Global variable @Cloud_username available

SeqLaz:
Listed on the available Global Variables:
image

Retrieving the @cloud_username
image


m-kuhn:
@SeqLaz thanks

- does this work when you restart qgis with auto login enabled?
- does this remove the variables if you log out?

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -1174,3 +1176,11 @@ def _on_logout_success(self) -> None:
     def _on_logout_failed(self, err: str) -> None:
         self.set_feedback("Sign out failed: {}".format(str(err)))
         self.avatarButton.setEnabled(True)
+
+    def set_variable_cloud_username(self) -> None:
+        QgsExpressionContextUtils.setGlobalVariable(
+            "cloud_username", self.network_manager.auth().config("username")
```
gounux, line 1182: Please note that it will not always be possible to use this `network_manager.auth().config("username")` method, when SSO login will be implemented and merged.

A user logged in with a SSO provider will likely not have the `username` in the stored `QgsAuthMethodConfig`, which will use OAuth2 method and not username/password like a `Basic` auth configuration.

[See the new function in its current WIP state here](https://github.com/opengisch/qfieldsync/blob/0acb2d7b6896843d92b2ca488a194534b0d79284/qfieldsync/core/cloud_api.py#L263), that could also change in the near future.

SeqLaz: That's true, and is great that there is a method already existing!.... Let me use it!

SeqLaz: When merged I will rebase it and use the method get_username

suricactus: You can rebase already and change the base branch of this PR, the approach in the other PR is here to stay.

SeqLaz: Great!.... Thanks @suricactus

SeqLaz:
Hey @suricactus @gounux by using the method `.get_username()` is working correctly on adding the variable @cloud_username!

qfieldsync/gui/cloud_projects_dialog.py
SeqLaz: Thanks for the method @gounux!

Issue #658, opened 2025-04-10
Map Themes cause AttributeError: 'QgsMapLayerComboBox' object has no attribute 'setProject'


byteit101:
# Steps to reproduce

 * Use Debian 12's QGIS 3.22
 * Install the latest qfieldsync
 * In the layers panel, add a theme (no layers or group required, but this also works with a layer or group)
 * click `Configure current project` in QField toolbar
 * The UI looks funky and an error is produced:

```
2025-04-09T21:31:24     WARNING    Traceback (most recent call last):
File "/home/byteit101/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/qfield_sync.py", line 63, in createWidget
return ProjectConfigurationWidget(parent)
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/home/byteit101/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/project_configuration_widget.py", line 127, in __init__
self.reloadProject()
File "/home/byteit101/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/project_configuration_widget.py", line 170, in reloadProject
self.mapThemesConfigWidget = MapThemesConfigWidget(
^^^^^^^^^^^^^^^^^^^^^^
File "/home/byteit101/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/mapthemes_config_widget.py", line 43, in __init__
self.reload(configuration)
File "/home/byteit101/.local/share/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/gui/mapthemes_config_widget.py", line 62, in reload
cmb.setProject(self.project)
^^^^^^^^^^^^^^
AttributeError: 'QgsMapLayerComboBox' object has no attribute 'setProject'

```

# Versions

QGIS version | 3.22.16-Białowieża | QGIS code branch | Release 3.22
-- | -- | -- | --
Qt version | 5.15.8
Python version | 3.11.1
GDAL/OGR version | 3.6.2
PROJ version | 9.1.1
EPSG Registry database version | v10.076 (2022-08-31)
GEOS version | 3.11.1-CAPI-1.17.1
SQLite version | 3.40.1
PostgreSQL client version | 15.1 (Debian 15.1-1+b1)
SpatiaLite version | 5.0.1
QWT version | 6.1.4
QScintilla2 version | 2.13.3
[...]

PR #659, opened 2025-04-11
Fix QGIS 3.22 compatibility issue

nirvn:
Fixes https://github.com/opengisch/qfieldsync/issues/658


suricactus (approved):
Is it possible that we add a version check branching instead of entirely deleting the line? In theory having the project being set is slightly more future proof in the context of multiple projects available at a time.

I am -0 to merge like this too.

nirvn:
@suricactus , we use the same widget _without_ the setProject elsewhere.

nirvn:
@suricactus , changed for all 3 usages of the widget.

m-kuhn:
Should we release and bump the minimum qgis version to something more recent?

suricactus:
I don't think so since it does not require any significant maintenence for now. It is the default current Debian version too.

PR #660, opened 2025-04-11
Fix attachment directories list widget's tiny height issue

nirvn:
The attachment directories list widget's height can get really small - ie _too_ small - when expanding a couple of group boxes in there. 

While the whole project configuration panel deserves a re-styling, this PR provides a nice band-aid by adding a minimum height so the list remains usable.


nirvn:
@suricactus 

image

PR #662, opened 2025-04-13
Cloud project transfer dialog bug hunting and improvements

nirvn:
This mainly started with fixing something that's been a problem with QFieldSync for a long while and negatively impacts the process of uploading projects to QFieldCloud: the feedback table widget shown when layer(s) or the project being uploaded has issues will often appear *twice*. 

Here's an example:

Screenshot From 2025-04-13 11-23-31

The table is already scary, we just end up scaring people twice as much due to this glitch ;)

While looking into this dialog, I then took the time to improve the vertical spread of widgets there. ATM, none of the widgets expand, leaving empty spaces all around:

Screenshot From 2025-04-13 11-42-36

By having the feedback table expand to fill all available space, we end up with a much nicer dialog.

Finally, I've removed the use of ALL CAPS for the warning and error label, added nice icons (to better match the feedback dialog QGIS has within the layout designer), and I've merged the source + message columns to better manage the space we have.

The final look:

image


qfieldsync/gui/checker_feedback_table.py
```diff
@@ -54,7 +45,16 @@ def __init__(self, checker_feedback: ProjectCheckerFeedback, *args, **kwargs):
                 self.insertRow(row)
 
                 # first column
-                item = QTableWidgetItem(str(feedback.level.name))
+                if feedback.level == Feedback.Level.WARNING:
+                    level_icon = "/mIconWarning.svg"
+                    level_text = self.tr("Warning")
```
m-kuhn, line 50: Is this even needed?

qfieldsync/gui/checker_feedback_table.py
```diff
@@ -54,7 +45,16 @@ def __init__(self, checker_feedback: ProjectCheckerFeedback, *args, **kwargs):
...
+                else:
+                    level_icon = "/mIconCritical.svg"
+                    level_text = self.tr("Error")
+
+                item = QTableWidgetItem(
+                    QgsApplication.getThemeIcon(level_icon), level_text
```
m-kuhn, line 56: I wonder if we want to make the text just a tooltip, I don't see us having any benefit of this in written form here

qfieldsync/ui/cloud_transfer_dialog.ui
```diff
@@ -95,8 +95,14 @@
...
         <widget class="QLabel" name="feedbackLabel_2">
+         <property name="sizePolicy">
+          <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
+           <horstretch>0</horstretch>
+           <verstretch>0</verstretch>
+          </sizepolicy>
+         </property>
          <property name="text">
-          <string>The project check provides automatic feedback for some common mistakes when using QField, QFieldSync and QFieldCloud. Lines starting with ERROR show problems in your project configuration that must be fixed before proceeding. Lines starting with WARNING are suggestions what you can improve.</string>
+          <string>The project check provides automatic feedback for some common mistakes when using QField, QFieldSync and QFieldCloud. Errors show problems in your project configuration that must be fixed before proceeding. Warnings are suggestions what you can improve.</string>
```
m-kuhn, line 105: Should we use something similar to a :bulb: instead of the warning sign? Would look much friendlier.

m-kuhn (approved):
Thanks for these fixes, mergable with or without adjustments as commented

qfieldsync/gui/checker_feedback_table.py
nirvn: One benefit is that it ties in with the explainer above. I'd be keen to see it stay.

qfieldsync/ui/cloud_transfer_dialog.ui
nirvn: @m-kuhn , big +1 , I like it. I'll implement

nirvn:
@m-kuhn , here's a bulb:

image

nirvn:
Other improvements we could do (on the libqfieldsync side of things): remove exclamation marks (!) in our strings.

m-kuhn:
Could we also remove the word "Warning" and turn it into a tooltip?

nirvn:
@m-kuhn , hmm, well, without the explicit word there, I feel the description paragraph above becomes a bit more obscure no?

m-kuhn:
I am not attached to the paragraph. Are you?

nirvn:
@m-kuhn , no -- what about this:

_The project checks provides automatic feedback for some common mistakes when preparing projects for QField and QFieldCloud. Find below a list of improvement suggestions and errors that must be fixed before proceeding._ 

?

nirvn:
@m-kuhn , @suricactus , done as you suggested:

image

qfieldsync/gui/checker_feedback_table.py
```diff
@@ -21,31 +21,24 @@
...
 class CheckerFeedbackTable(QTableWidget):
     def __init__(self, checker_feedback: ProjectCheckerFeedback, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self.setColumnCount(3)
-        self.setHorizontalHeaderLabels(
-            [self.tr("Level"), self.tr("Source"), self.tr("Message")]
-        )
+        self.setColumnCount(2)
+        self.setHorizontalHeaderLabels(["  ", self.tr("Message")])
```
suricactus, line 37: Is there a reason why the column header is two empty spaces? If yes, I guess it would be nice to add a small comment.

qfieldsync/gui/checker_feedback_table.py
```diff
@@ -64,16 +65,12 @@ def __init__(self, checker_feedback: ProjectCheckerFeedback, *args, **kwargs):
...
-
-                # third column
                 item = QTableWidgetItem()
                 item.setFlags(Qt.ItemIsEnabled)
-                self.setItem(row, 2, item)
+                item.setToolTip(level_text)
+                self.setItem(row, 1, item)
 
-                label = QLabel(feedback.message)
+                label = QLabel("**{}**\n\n{}".format(source, feedback.message))
```
suricactus, line 73:
```suggestion
                # we do not escape the values on purpose to support Markdown/HTML
                label = QLabel("**{}**\n\n{}".format(source, feedback.message))
```

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -227,12 +227,13 @@ def show_project_compatibility_page(self):
         if feedback and feedback.count > 0:
             has_errors = len(feedback.error_feedbacks) > 0
 
-            feedback_table = CheckerFeedbackTable(feedback)
-            self.feedbackTableWrapperLayout.addWidget(feedback_table)
-            self.stackedWidget.setCurrentWidget(self.projectCompatibilityPage)
-            self.buttonBox.button(QDialogButtonBox.Apply).setVisible(True)
-            self.buttonBox.button(QDialogButtonBox.Apply).setEnabled(not has_errors)
-            self.buttonBox.button(QDialogButtonBox.Apply).setText(self.tr("Next"))
+            if self.feedbackTableWrapperLayout.count() == 0:
```
suricactus, line 230: I do not follow when `feedbackTableWrapperLayout.count()` is (not) 0? And therefore I don't know if we do (not) need an `else`?

nirvn: The showEvent can be triggered more than once when showing a dialog, hence why we had this bad issue with the feedback widget being duplicated. 

The check here insures we only add it once

qfieldsync/gui/checker_feedback_table.py
nirvn: Just meant as a way to insure a minimum width really. It can be empty too

qfieldsync/gui/cloud_transfer_dialog.py
suricactus: Great! Can we add a comment mentioning this?

suricactus: I guess we can also move the `has_errors` in the `if` block too

qfieldsync/gui/checker_feedback_table.py
suricactus: If we need the minimum width I am ok, just mention this in a comment, because it is not immediately clear.

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: @suricactus , we can move the whole function behind that condition.

PR #663, opened 2025-04-13
Cloud project transfer dialog bug hunting and improvements, part 2

nirvn:
More polishing and improvements to the cloud project transfer dialog, this time on the panel that has users pick which files are uploaded/download to/from the cloud:

- on dark themes, the local project directory text edit widget was looking bad, the PR fixes that by avoiding hardcoded stylesheet colors
- the project hyperlink that's meant to bring users to the project page on the web did not work, fixed
- the local project directory didn't offer much at all in terms of UI/UX, a button has been added so users can open it directly

Beyond that, I've re-ordered the widgets a bit so the panel focuses on the job at hand, i.e. to transfer files. The project's hyperlink to the web app (the project name being already int the dialog title bar) and the local directory text edit widgets are secondary details/actions that should hit the eye first. As such, I've moved these to sit below the transfer widgets section, and into a collapsible groupbox (expanded by default, until users decide to save space :) ).

Before:
Screenshot From 2025-04-13 13-31-44

PR, with project details expanded:
Screenshot From 2025-04-13 13-28-08

PR, with project details collapsed:
Screenshot From 2025-04-13 13-28-23

### DIFF
--- a/qfieldsync/gui/cloud_transfer_dialog.py
+++ b/qfieldsync/gui/cloud_transfer_dialog.py
@@ -30,7 +30,7 @@ from libqfieldsync.offline_converter import ExportType
 from libqfieldsync.project_checker import ProjectChecker
 from libqfieldsync.utils.file_utils import get_unique_empty_dirname
 from libqfieldsync.utils.qgis import get_qgis_files_within_dir
-from qgis.core import QgsProject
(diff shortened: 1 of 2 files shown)


m-kuhn (approved):
Good step in the right direction

PR #666, opened 2025-05-02
Seamless localized datasets upload support

nirvn:
This PR takes on the work done by @tdrivas and finishes the handling of localized datasets upload when synchronizing a cloud project.

The following additions are in the PR:
- we do not rely on QgsProject.instance() to get localized datasets but instead we do it by temporarily opening a locally stored cloud project file to fetch them
- the upload relies on the ThrottledFileTransfer class, which gives us UI feedback and control over how many files we upload at the same time
- we rely on project cache to avoid hard-coded project fetch
- a minimal UI was added to allow for users to skip localized dataset upload (see the screenshot below):

image

I've been re-thinking about this, and IMHO, we should actually avoid uploading a localized dataset when it's present remotely but has a different checksums. That should be handled when someone manually synchronize the localized_dataset project itself.

Thinks that could be improved in a follow up PR (but really not that required for an initial drop):
- adding a warning message for checksum mismatch

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -735,7 +735,7 @@ class CloudNetworkAccessManager(QObject):
             self._nam.cookieJar().deleteCookie(cookie)
 
 
-    def ensure_localized_dataset_project(self) -> Optional[str]:
+    def get_localized_datasets_project(self) -> Optional[str]:
         """
         Ensures that the 'localized_datasets' project exists under the user's organization.
 
@@ -745,33 +745,22 @@ class CloudNetworkAccessManager(QObject):
         try:
          
           
-            org_username = self.user_details.get("username")
+            username = self.user_details.get("username")
 
-            if not org_username:
-                QgsMessageLog.logMessage("User not authenticated", "QFieldSync", Qgis.Warning)
-                return None
-            
             existing_projects = self.get_projects_not_async()
 
             for project in existing_projects:
-                if project.get("name") == "localized_datasets":
-                    QgsMessageLog.logMessage(
-                        "'localized_datasets' project already exists."+project.get("id"),
-                        "QFieldSync",
-                        Qgis.Info,
-                    )
+                if project.get("name") == "localized_datasets" and project.get("owner") == username:
                     return project
 
             reply = self.create_project(
                 name="localized_datasets",
-                owner=org_username,
+                owner=username,
                 description="Localized datasets for QField",
                 private=True,
             )
             
             project_data = self.json_object(reply)
-
-            QgsMessageLog.logMessage("Project ID:"+project_data.id, "QFieldSync", Qgis.Warning)
 
             return project_data
 
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -735,28 +735,33 @@ class CloudNetworkAccessManager(QObject):
             self._nam.cookieJar().deleteCookie(cookie)
 
 
-    def get_localized_datasets_project(self) -> Optional[str]:
+    def get_localized_datasets_project(self, owner: str) -> Optional[str]:
         """
-        Ensures that the 'localized_datasets' project exists under the user's organization.
+        Retrieve the 'localized_datasets' project ID for a given owner.
+
+        This ensures that the 'localized_datasets' project is retrieved for the specified owner,
+        typically an organization or user that owns the main project. If such a project does not exist,
+        None is returned.
+
+        Args:
+            owner (str): The username of the project owner (person or organization).
 
         Returns:
-            str | None: The project ID of the 'localized_datasets' project, or None if creation failed.
+            Optional[str]: The ID of the 'localized_datasets' project if found, otherwise None.
         """
         try:
          
           
-            username = self.user_details.get("username")
-
             existing_projects = self.get_projects_not_async()
 
             for project in existing_projects:
-                if project.get("name") == "localized_datasets" and project.get("owner") == username:
+                if project.get("name") == "localized_datasets" and project.get("owner") == owner:
                     return project
 
             reply = self.create_project(
                 name="localized_datasets",
-                owner=username,
-                description="Localized datasets for QField",
+                owner=owner,
+                description="Localized datasets for QField(Sync)",
                 private=True,
             )
             
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -986,3 +986,32 @@ class CloudProjectsCache(QObject):
         for project in self._projects:
             if dirpath == project.local_dir:
                 project.refresh_files()
+
+    def ensure_localized_dataset_project(self) -> Optional[str]:
+        """
+        Ensures that the 'localized_datasets' project exists under the user's organization.
+
+        Returns:
+            str | None: The project ID of the 'localized_datasets' project, or None if creation failed.
+        """
+        try:
+            for project in self.network_manager.projects_cache.projects():
+                if project["name"] == "localized_datasets":
+                    return project["id"]
+
+            account = self.network_manager.account()
+
+            if not account or not account.get("is_organization_owner"):
+                return None
+
+            org_username = account["username"]
+            project_data = self.network_manager.create_project(
+                name="localized_datasets",
+                owner_username=org_username,
+                private=True,
+            )
+            
+            return project_data["id"]
+
+        except Exception as err:
+            self.project_files_error.emit(project_id, str(err))
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -37,7 +37,7 @@ from qgis.core import (
     QgsProject,
     QgsMessageLog
 )
-from qgis.PyQt.QtCore import QFileSystemWatcher, QObject, QUrl, QUrlQuery, pyqtSignal
+from qgis.PyQt.QtCore import QEventLoop, QFileSystemWatcher, QObject, QUrl, QUrlQuery, pyqtSignal
 from qgis.PyQt.QtNetwork import (
     QHttpMultiPart,
     QHttpPart,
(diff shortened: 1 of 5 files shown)


qfieldsync/core/cloud_api.py
```diff
@@ -728,6 +736,56 @@ def get_last_login_error(self) -> str:
 
         return error_str
 
+    def get_localized_datasets_project(self, owner: str) -> Optional[str]:
```
suricactus, line 739:
```suggestion
    def get_or_create_localized_datasets_project(self, owner: str) -> Optional[str]:
```

qfieldsync/core/cloud_transferrer.py
```diff
@@ -623,7 +717,9 @@ def __init__(
                 self.cloud_project,
                 self.transfer_type,
                 file,
-                self.temp_dir.joinpath(str(self.transfer_type.value), file.name),
+                Path(file.local_dir).joinpath(file.name)
+                if use_file_local_dir
+                else self.temp_dir.joinpath(str(self.transfer_type.value), file.name),
             )
```
suricactus, line 723: Can we move this to a variable, it's unreadable as a parameter?

Also, what if `temp_dir` is `None`?

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -310,6 +316,34 @@ def show_end_page(
...
+                        localized_datasets_project.user_role == "admin"
+                        or localized_datasets_project.user_role == "manager"
+                    ):
+                        self.localized_datasets_project = localized_datasets_project
+                        self.localized_datasets_files = localized_datasets_files
+                        reply = self.network_manager.projects_cache.get_project_files(
+                            localized_datasets_project.id
+                        )
+                        reply.finished.connect(lambda: self.prepare_project_transfer())
+                        return
```
suricactus, line 344: I am wondering if it is possible to write it without having 4 levels of nested `if`s?

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -323,7 +357,38 @@ def prepare_project_transfer(self):
...
-        if len(list(self.cloud_project.files_to_sync)) == 0:
+        # Remove files that are already present remotely, override should be done manually
+        if self.localized_datasets_project and self.localized_datasets_files:
+            localized_data_paths = (
+                QgsApplication.instance().localizedDataPathRegistry().paths()
+            )
+            localized_datasets_project_files = (
+                self.localized_datasets_project.get_files()
+            )
+            localized_datasets_project_names = []
```
suricactus, line 368:
```suggestion
            localized_datasets_project_filenames = []
```
I guess those are filenames, no?

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -570,7 +656,12 @@ def _start_synchronization(self):
             assert self.project_transfer
 
             self.project_transfer.sync(
-                files["to_upload"], files["to_download"], files["to_delete"]
+                files["to_upload"],
+                files["to_download"],
+                files["to_delete"],
+                self.localized_datasets_files
+                if self.uploadLocalizedDatasetsCheck.isChecked()
+                else [],
```
suricactus, line 664: May we move this to a separate variable instead of abusing the inline if's in the argument list :)

suricactus (commented):
Great addition, thank you! 

A few comments, mostly about readability. The main problem I see here is if we can prepare the `localized_datasets_files` to be converted to a list of project files, so we keep the datastructures similar enough around the code.

qfieldsync/core/cloud_transferrer.py
```diff
@@ -702,13 +798,14 @@ def __init__(
         self.filename_to_index: Dict[str, int] = {}
 
         for transferrer in transferrers:
-            for filename, transfer in transferrer.transfers.items():
-                self.filename_to_index[filename] = len(self.transfers)
-                self.transfers.append(transfer)
+            if transferrer:
```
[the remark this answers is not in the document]
nirvn: @suricactus , when we have no localized datasets to upload, the throttled_localized_datasets_uploader is None.

qfieldsync/core/cloud_transferrer.py
nirvn: Fixed, it'll never be None (like it used to be)

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: One level removed.

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: @suricactus , no, they are ProjectFile.name (i.e. the relative name string to the root of the local dir)

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: Note to self: Ivan doesn't like inline if's :wink:

nirvn:
@suricactus , review addressed.

qfieldsync/core/cloud_project.py
```diff
@@ -390,6 +394,54 @@ def local_project_file(self) -> Optional[ProjectFile]:
...
+            if Qgis.versionInt() >= 32600:
+                read_flags |= QgsProject.FlagDontLoad3DViews
+
+            temporary_project = QgsProject()
+            temporary_project.read(str(self.local_project_file.local_path), read_flags)
+            layers = temporary_project.mapLayers()
+
+            localized_datasets_files = []
+            for layer_id, layer in layers.items():
+                if layer.dataProvider():
```
suricactus, line 419:
```suggestion
                if not layer.dataProvider():
                    continue
```

qfieldsync/core/cloud_project.py
```diff
@@ -390,6 +394,54 @@ def local_project_file(self) -> Optional[ProjectFile]:
...
+                    metadata = QgsProviderRegistry.instance().providerMetadata(
+                        layer.dataProvider().name()
+                    )
+                    metadata.decodeUri(layer.source())
+
+                    filename = metadata.decodeUri(layer.source()).get("path", "")
+                    if filename:
```
suricactus, line 426:
```suggestion
                    if not filename:
                        continue
```

qfieldsync/core/cloud_project.py
```diff
@@ -390,6 +394,54 @@ def local_project_file(self) -> Optional[ProjectFile]:
...
+                        for localized_data_path in localized_data_paths:
+                            if filename.startswith(localized_data_path):
```
suricactus, line 428:
```suggestion
                            if not filename.startswith(localized_data_path):
                                continue
```

qfieldsync/core/cloud_project.py
```diff
@@ -390,6 +394,54 @@ def local_project_file(self) -> Optional[ProjectFile]:
...
+                                localized_datasets_files.append(
+                                    ProjectFile(
+                                        {"name": localized_datasets_name},
+                                        local_dir=localized_data_path,
+                                    )
+                                )
+
+            return localized_datasets_files
+        else:
+            return []
```
suricactus, line 443: We can make an early return here, but it is a funky condition.

qfieldsync/core/cloud_transferrer.py
```diff
@@ -618,12 +693,19 @@ def __init__(
         self.transfer_type = transfer_type
 
         for file in self.files:
+            if use_file_local_dir:
+                destination = Path(file.local_dir).joinpath(file.name)
+            elif self.temp_dir:
```
suricactus, line 698: I guess there should be an assert in the `else` part in case the `if` and `elif` are not met?

qfieldsync/core/cloud_transferrer.py
```diff
@@ -398,6 +451,27 @@ def import_qfield_project(self) -> bool:
     def _on_logout_success(self) -> None:
         self.abort_requests()
 
+    def _upload_localized_dataset_files(self) -> None:
+        # nothing to upload
+        if len(self._files_to_upload_for_localized_datasets) == 0:
+            self.localized_datasets_upload_progress.emit(1)
+            self.localized_datasets_upload_finished.emit()
+            # NOTE check _on_localized_datasets_upload_finished
```
suricactus, line 459:
```suggestion
            # NOTE the execution logic here continues in the `_on_localized_datasets_upload_finished`, as now we have to upload the regular project files.
```

suricactus (approved):
Absolutely great evolution, I feel we greatly simplified what we are doing now.

A few code style suggestion, I would suggest we use early returns whenever possible instead of nested `if`s. And the other one is to be paranoid with `elif`s , so we get early exceptions.

Anyways, it is already mergeable.

qfieldsync/core/cloud_transferrer.py
nirvn: That's an artifact from previous logic, it should just be else: here, changed.

qfieldsync/core/cloud_transferrer.py
```diff
@@ -163,16 +176,36 @@ def sync(
...
+        ):
+            self.throttled_localized_datasets_uploader = ThrottledFileTransferrer(
+                self.network_manager,
+                self.localized_datasets_project,
+                self._files_to_upload_for_localized_datasets,
+                FileTransfer.Type.UPLOAD,
+                use_file_local_dir=True,
+            )
+        else:
+            self.throttled_localized_datasets_uploader = None
```
suricactus, line 196:
```suggestion
        // set the `_files_to_upload_for_localized_datasets` only if there is a localized project available
        if self.localized_datasets_project:
            for localized_file in files_to_upload_for_localized_datasets:
                self._files_to_upload_for_localized_datasets[str(project_file.path.as_posix())] = localized_file
        else:
            self._files_to_upload_for_localized_datasets =  {}

        self.throttled_uploader_for_localized_datasets = ThrottledFileTransferrer(
            self.network_manager,
            self.localized_datasets_project,
            self._files_to_upload_for_localized_datasets.values(),
            FileTransfer.Type.UPLOAD,
            use_file_local_dir=True,
        )
```


- Make sure the `_files_to_upload_for_localized_datasets` is aligned as data type with the rest of the `_files_to` datatypes.
- Renamed `throttled_localized_datasets_uploader` to `throttled_uploader_for_localized_datasets` to match the naming with the `_files_to_upload_for_localized_datasets`.
- Made the `throttled_uploader_for_localized_datasets` always available, even though it might be empty.
- Addressed the latest comment in https://github.com/opengisch/qfieldsync/pull/666/files#r2074194109 , so we reduce the code branching outside this function.

suricactus (commented):
Final review I guess. Spot a few typing improvements I will do in a follow-up PR.

PR #668, opened 2025-05-08
Fixing missing argument files_to_upload_for_localized_datasets on the cloud transferrer

SeqLaz:
This PR fix:
image


suricactus (approved):
Good catch, thanks!

PR #678, opened 2025-06-11
fix: on windows the newly created fs directories might not be deletable

suricactus:
Apparently the passed `mode` value on `os.mkdir` and `Path.mkdir` is not respected by Windows prior to 3.13. What is more, in 3.13 Windows will handle only 0o700, the rest of the values will be ignored.


nirvn (approved):

suricactus:
Merged due to lack of feedback.

PR #682, opened 2025-06-21
Add additional default attachment directories to cover all default attachment expressions

nirvn:
We should have done this a long time ago -- once upon a time, when we added audio, video, and other file attachments support, the default expression has files put in more than a single DCIM directory. If the users to not add these extra folders, they will end up not able to package attachments when downloading the project on QField devices (i.e. https://github.com/opengisch/QField/issues/6409)


suricactus (approved):

vlebert:
Good idea, but can we remove it ? Sometimes it may be useful not to share all local pictures for lighter packaging

nirvn:
@vlebert , yes, they are there by default but can be removed

PR #685, opened 2025-06-28
Project properties' attachment directories UI revamp

nirvn:
This PR revamps the attachment directories UI found in the project properties dialog, and adds a new "data" directory type. Here's how it looks:

image

Improvements include:
- A clear way to add new directories (i.e., users don't need to guess that double clicking on a blank row is the way you add a row :))
- A clear way to remove directories

The need for a new data directory type emerged since we added on-demand attachment download in QField/QFieldCloud. If attachments are downloaded on demand, our project plugin / symbology / decoration assets placed in a subfolder still needs to be downloaded. That will now happen with data directories.

### DIFF
--- a/qfieldsync/core/preferences.py
+++ b/qfieldsync/core/preferences.py
@@ -31,6 +31,7 @@ class Preferences(SettingManager):
                 "attachmentDirs", Scope.Project, ["DCIM", "audio", "video", "files"]
             )
         )
+        self.add_setting(Stringlist("dataDirs", Scope.Project, []))
         self.add_setting(Dictionary("qfieldCloudProjectLocalDirs", Scope.Global, {}))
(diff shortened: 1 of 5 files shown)


suricactus (approved):
Code looks good.

Two questions that do not require an answer:
- I see why you went with "data" as a name of the rest of the dirs, not sure if it is the best name though. Unfortunately I cannot come up with something better myself...
- I extensively use tabs to navigate forms, not sure how the removed tabstop entry affects this.

Feel free to merge with or without answering these two.

nirvn:
@suricactus , I also agree data isn't great but couldn't think of a better name either. 

I'll check on the tab stop situation today, good reminder.

nirvn:
@suricactus , we also need this in: https://github.com/opengisch/QFieldCloud/pull/1271

PR #686, opened 2025-06-29
Layers configuration widget cleanup + map themes table widget harmonization

nirvn:
The PR attempts at cleaning up the layers configuration widget to make it less intimidating and feel less crowded.

Before:
Screenshot From 2025-06-29 09-42-24

PR:
Screenshot From 2025-06-29 09-40-29

Improvements include:
- Removing the vertical header (i.e. 1,2,3,4,5,etc.), it just takes up space for no reason since the numbers are not used here
- Getting rid of the repeated [ Properties ] push button in favor of a much smaller tool button with a simple settings gear icon which results in more space for the layer label and a less intimidating view IMHO
- Alternating row colors, because it's a nice, subtle guide

The map themes table widget got the same visual tweaks (no vertical header, alt. row color)

### DIFF
--- a/qfieldsync/gui/layers_config_widget.py
+++ b/qfieldsync/gui/layers_config_widget.py
@@ -24,14 +24,14 @@ import os
 from typing import Callable
 
 from libqfieldsync.layer import LayerSource, SyncAction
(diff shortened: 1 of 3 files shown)


suricactus (approved):
Great improvements, thanks!

PR #689, opened 2025-07-06
Project properties restructuring and UX improvements

nirvn:
_Preface: the motivation for this PR comes as I'm about to introduce another attachment-related project configuration into QFieldSync and I did *not* want to simply throw another setting in the advanced settings groupbox, which had become an aggregation of everything under the sun._

This PR focuses on restructuring the QFS project properties widget, as well as add short introduction labels to sections to provide preliminary guidance to users. This adopts a practice increasingly used in QGIS itself and IMHO can be quite useful in better sharing the purpose of various settings.

Before (left) vs. PR (right):

image322

Improvements here include:
- the regrouping of AOI and base map together; these settings work together and it was really not that clear before now
- an explicit, visual guide as to what it means to leave an empty AOI
- crucial help in guiding people on how to set the right packaging action

On the regrouping of attachment settings into its own tab, here's a before( left) vs. PR (right):

image3462

In a few days, that section will have another setting (photo information stamping customization), and will look ever nicer :sunglasses: 

I've also taken the time to move the semi-hidden but crucial mouse hover UUID tip into what I believe is a better context, within the layer properties panel alongside the relationship settings (where UUID usage is most important). That was a good occasion to apply a similar cleanup.

Before (left) vs. PR (right):

image374

By moving attachments and relationship setting into its own collapsible groupbox, it allows us with a simple logic in the code to collapse these when no attachment editor widgets and/or relationships exist for a given vector layer, which leaves the space for other settings including the tracker session groupbox. Here's an example of what it looks like to open a vector layer with a tracker session configured where no attachments and relationships exist:

image398

### DIFF
--- a/qfieldsync/gui/map_layer_config_widget.py
+++ b/qfieldsync/gui/map_layer_config_widget.py
@@ -25,7 +25,6 @@ import os
 from libqfieldsync.layer import LayerSource
 from qgis.core import QgsMapLayer, QgsProject, QgsProperty, QgsPropertyDefinition
 from qgis.gui import QgsMapLayerConfigWidget, QgsMapLayerConfigWidgetFactory, QgsSpinBox
-from qgis.PyQt.QtWidgets import QLabel
 from qgis.PyQt.uic import loadUiType
 
 from qfieldsync.core.message_bus import message_bus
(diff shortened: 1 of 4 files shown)


itsakifa:
@nirvn very welcome changes! I have a few questions:
- when defining an AOI, I would assume that it is automatically taken into account for offlining. I find it a bit confusing to have an additional checkbox (I just noticed that, even if it was already there before) to make it happen. What is the difference between defining an AOI + not checking the checkbox versus leaving the AOI empty to "use the full project extent"?
- Is the AOI and basemp setting taken into account for QFieldCloud? I thought this is for now only working with cable export.
- When in the attachment and document tab, I find it strange to still have the AOI and geofencing settings on the same page. -> If "QFC packaging" / "cable packaging" only influences the first group box (packaging actions), I would find it more logical to have a "General" tab for AOI, basemap, geofencing, attachments, ... and then a separate tab for QFC packaging actions and cable packaging actions. If other settings are dependent on either QFC or cable, then it should stay separated like this, but without repeating elements that have only to do with one or the other.

nirvn:
@itsakifa , thanks for the feedback.

- AOI drives two options: the filtering of features being offlined, and/**or** the extent of the generated base map. That's why you can define an AOI while not checking the filtering option.
- That's correct, ATM the basemap generation is a cable-only setting, although it is likely to change in the future (hence why I'm keen to not move it into a cable tab only to remove it afterwards). The intro label hints at that.
- The QFieldCloud packaging, Cable packaging, and Attachments & directories tab site on top of a scroll area under which more settings reside (i.e. geofencing for e.g.), they are not part of individual tabs.

itsakifa:
@nirvn thanks for the additional info. It was visually not so clear to me that the tab sections sit on top of the rest, but I see now 🙈

nirvn:
@suricactus , re 1 & 2, part of the motivation in adding one more tab is to try and better utilize the space. It's not ideal, but IMHO I think it's an improvement and to me it feels natural.

Re 3, glad you want these radio buttons away, I do too! :) I think we can address that in a follow up PR, happy to move it to your suggested alternative location.

Re 4, OK , I'll do it here before merging.

Re 5, I like the idea, I can execute in a separate PR.

nirvn:
@suricactus , here's point 4 addressed:

image

nirvn:
Since there was no stated blocker here, I'm gonna go ahead and merge to unlock more development on my end. I'm sure we'll continue improving things as we go along :)

PR #690, opened 2025-07-10
Add logic to avoid auto uploading offline editing-driven datasets to QFieldCloud

nirvn:
This pull request guards users from unknowingly overwrite datasets tied to layers set to offline editing when triggering a cloud transfer push/pull within QFieldSync. 

Instead of automatically setting the upload/download action for datasets tied to layers being offlined to _upload_ when the dataset's local modified time is more recent than what is on QFieldCloud, the cloud transfer dialog now defaults to _no action_. That is meant to prevent accidental upload and force the users to be intentional about the action for such datasets.

For non-datasets or datasets set to direct access, nothing changes. Similarly, if a dataset that is set to offline editing has a modification time that is more recent than the locally stored file, we are still auto-setting the action to download, nothing changes on that front so people won't start reporting "where's my data?!" :)

Here's what it looks like when the local version of my cloud project has its project file modified as well as a geopackage dataset modified (because WAL et al) that is configured to be offlined:

image

### DIFF
--- a/qfieldsync/gui/cloud_transfer_dialog.py
+++ b/qfieldsync/gui/cloud_transfer_dialog.py
@@ -517,6 +517,42 @@ class CloudTransferDialog(QDialog, CloudTransferDialogUi):
         # ##########
         stack = []
 
+        offline_layers_paths = self._get_offline_layers()
+
+        for project_file in self.project_transfer.cloud_project.files_to_sync:
+            parts = tuple(project_file.path.parts)
+            for part_idx, part in enumerate(parts):
+                if len(stack) > part_idx and stack[part_idx][0] == part:
+                    continue
+                else:
+                    stack = stack[0:part_idx]
+
+                item = QTreeWidgetItem()
+                item.setText(0, part)
+                item.setExpanded(True)
+
+                stack.append((part, item))
+
+                if len(stack) == 1:
+                    self.filesTree.addTopLevelItem(item)
+                else:
+                    stack[-2][1].addChild(stack[-1][1])
+
+                # the length of the stack and the parts is equal for file entries
+                if len(stack) == len(parts):
+                    item.setData(0, Qt.UserRole, project_file)
+                    is_offline_layer = (
+                        project_file.local_path_exists
+                        and str(project_file.local_path) in offline_layers_paths
+                    )
+                    self.add_file_checkbox_buttons(item, project_file, is_offline_layer)
+                else:
+                    # TODO make a fancy button that marks all the child items as checked or not
+                    pass
+        self.filesTree.expandAll()
+        # NOTE END algorithmic part
+
+    def _get_offline_layers(self):
         offline_layers_paths = []
         if self.cloud_project and self.cloud_project.is_current_qgis_project:
             project_layers = list(QgsProject.instance().mapLayers().values())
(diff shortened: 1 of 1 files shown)


qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -538,7 +569,12 @@ def build_files_tree(self):
                 # the length of the stack and the parts is equal for file entries
                 if len(stack) == len(parts):
                     item.setData(0, Qt.UserRole, project_file)
-                    self.add_file_checkbox_buttons(item, project_file)
+                    self.add_file_checkbox_buttons(
+                        item,
+                        project_file,
+                        project_file.local_path_exists
+                        and str(project_file.local_path) in offline_layers_paths,
+                    )
```
suricactus, line 577:
```suggestion
                    is_offline_layer = (
                        project_file.local_path_exists
                        and str(project_file.local_path) in offline_layers_paths
                    )

                    self.add_file_checkbox_buttons(item, project_file, is_offline_layer)
```

Took me a while to count the number of paratemers as 3 instead of 4, so I guess a good candidate to calculate the flag in a separate step.

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -516,6 +517,36 @@ def build_files_tree(self):
...
+            project.read(str(self.cloud_project.root_project_files[0]), read_flags)
+
+            project_layers = list(project.mapLayers().values())
+            for project_layer in project_layers:
+                provider_metadata = QgsProviderRegistry.instance().providerMetadata(
+                    project_layer.providerType()
+                )
+                metadata = provider_metadata.decodeUri(project_layer.source())
+                if metadata.get("path", "") != "":
+                    offline_layers_paths.append(metadata.get("path"))
```
suricactus, line 548:
```suggestion
        offline_layers_paths = self._get_offline_layers()
```

I think this is quite self-contained to become a separate function and improve readability.

qfieldsync/gui/cloud_transfer_dialog.py
suricactus: Didn't look too much in details, but I would assume we cannot add `LayerSource.is_offline_layer` flag that moves the logic where it makes most sense I guess?

suricactus (commented):
A few code readability comments, didn't look too much in the functionality. @gounux feel free to also review :)

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -538,13 +541,54 @@ def build_files_tree(self):
...
+                    )
+                    self.add_file_checkbox_buttons(item, project_file, is_offline_layer)
                 else:
                     # TODO make a fancy button that marks all the child items as checked or not
                     pass
         self.filesTree.expandAll()
         # NOTE END algorithmic part
 
+    def _get_offline_layers(self):
+        offline_layers_paths = []
```
gounux, line 556:
```suggestion
    def _get_offline_layers_paths(self) -> List[str]:
    """
    Get the paths of project layers configured for offline editing.
    """
        offline_layers_paths = []
```

Maybe a little docstring could be nice here. Feel free to reword / enhance / densify.

qfieldsync/gui/cloud_transfer_dialog.py
```diff
@@ -538,13 +541,54 @@ def build_files_tree(self):
...
+        elif self.cloud_project and len(self.cloud_project.root_project_files) == 1:
+            project = QgsProject()
+
+            read_flags = QgsProject.ReadFlags()
+            read_flags |= QgsProject.FlagDontResolveLayers
+            read_flags |= QgsProject.FlagDontLoadLayouts
+            if Qgis.versionInt() >= 32600:
+                read_flags |= QgsProject.FlagDontLoad3DViews
+
+            project.read(str(self.cloud_project.root_project_files[0]), read_flags)
```
gounux, line 575: Are we sure the file here is going to be a QGIS project file ?

gounux (commented):
Thanks ! Nothing much to say, glad to get the mind & hands into this part of qfieldsync's codebase.

I believe this will help users and reduce the number of annoying false manipulations they have to make up for, nice feature !

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: @gounux , yes, see https://github.com/opengisch/libqfieldsync/blob/master/libqfieldsync/utils/qgis.py#L69-L71

qfieldsync/gui/cloud_transfer_dialog.py
nirvn: @gounux , done.

suricactus (approved):
Thanks. Feel free to merge.

PR #712, opened 2025-08-13
Changing the Radio buttons to QToolButton

SeqLaz:
This is how it looks:
| Before | After |
| --- | --- |
|  |  |


duke-nyuki:
Task linked: [QF-6461 Convert the radio buttons for choosing "prefer online" or "prefer offline" layer into a button](https://app.clickup.com/t/2192114/QF-6461)

suricactus:
Great, can you add also the old/new comparison, so we can admire this change even more?

Would it be possible to add all/most of the new button within the `layers_config_widget.ui` instead of python code?

suricactus (approved):

suricactus:
Thanks, looks much better!

PR #715, opened 2025-08-16
Add a nifty search bar to filter cloud projects table

nirvn:
Because at one point, there are just too many projects not to want to have access to a search filter:

https://github.com/user-attachments/assets/8695aaa1-7949-4647-8a30-714141a10c6a

It also harmonizes the list with that offered on app.qfield.cloud as well as QField, both of which can be searched / filtered.


suricactus (approved):
Nice addition!

Issue #704, opened 2025-08-06
Cannot package attached files if QGIS and export directory are the same


SGroe:
### What is the bug or the crash? What were your expectations and what actually happened?

The following Python error occurs when the project is packaged to the directory where the QGIS project is stored.

```
2025-08-06T10:44:32     WARNING    Traceback (most recent call last):
              File "C:\Users/GIS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qfieldsync\gui\package_dialog.py", line 260, in package_project
              raise err
              File "C:\Users/GIS/AppData/Roaming/QGIS/QGIS3\profiles\default/python/plugins\qfieldsync\gui\package_dialog.py", line 253, in package_project
              self._offline_convertor.convert()
              File "C:\Users\GIS\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\qfieldsync\libqfieldsync_0d143dd958c38595bca8ade31293521182f7a3a2.whl\libqfieldsync\offline_converter.py", line 164, in convert
              self._convert(project)
              File "C:\Users\GIS\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\qfieldsync\libqfieldsync_0d143dd958c38595bca8ade31293521182f7a3a2.whl\libqfieldsync\offline_converter.py", line 374, in _convert
              copy_attachments(
              File "C:\Users\GIS\AppData\Roaming\QGIS\QGIS3\profiles\default\python\plugins\qfieldsync\libqfieldsync_0d143dd958c38595bca8ade31293521182f7a3a2.whl\libqfieldsync\utils\file_utils.py", line 158, in copy_attachments
              shutil.copyfile(filename, dest_filename)
              File "C:\OSGeo4W\apps\Python312\Lib\shutil.py", line 240, in copyfile
              raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
             shutil.SameFileError: WindowsPath('G:/xxx/Dokumente/ktn_gis/xxx/.fonts/BEV_DKM_Symbole_05_2012.ttf') and WindowsPath('G:/xxx/Dokumente/ktn_gis/xxx/.fonts/BEV_DKM_Symbole_05_2012.ttf') are the same file
```

Usually, I select another directory as destination. I recommend to either force to copy to another directory, or maybe just ignore the exception.

### Steps to reproduce the issue

1. Package project (with attached files) to directory where the QGIS project is stored

### QGIS version

3.40.9

[...]

PR #716, opened 2025-08-21
Prevent choosing the same directory as packaging destination for QFieldSync cable packaging

SeqLaz:
This PR fixes https://github.com/opengisch/qfieldsync/issues/704

This is how looks:

!

### DIFF
--- a/qfieldsync/gui/package_dialog.py
+++ b/qfieldsync/gui/package_dialog.py
@@ -213,6 +213,28 @@ class PackageDialog(QDialog, DialogUi):
         self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(False)
 
         packaged_project_file = Path(self.packagedProjectFileWidget.filePath())
+        packaged_project_dir = packaged_project_file.parent
+        open_project_path = self.project.fileName()
+        open_project_dir = Path(open_project_path).parent
+
+        if open_project_dir.resolve() == packaged_project_dir.resolve():
+            QMessageBox.critical(
+                self,
+                self.tr("Invalid Export Directory"),
+                self.tr(
+                    "The export directory cannot be the same as the source project directory.\n"
+                    "Please choose a different folder for the packaged project."
+                ),
+            )
+            return
+            self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(
+                False
+            )
+        else:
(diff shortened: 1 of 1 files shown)


qfieldsync/gui/package_dialog.py
```diff
@@ -213,6 +213,28 @@ def package_project(self):
...
+                self.tr("Invalid Export Directory"),
+                self.tr(
+                    "The export directory cannot be the same as the source project directory.\n"
+                    "Please choose a different folder for the packaged project."
+                ),
+            )
+            return
+            self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(
+                False
+            )
```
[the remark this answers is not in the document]
SeqLaz: Learned and noted!... Many thanks @suricactus

qfieldsync/gui/package_dialog.py
```diff
@@ -211,6 +211,28 @@ def package_project(self):
...
+                self,
+                self.tr("Invalid Export Directory"),
+                self.tr(
+                    "The export directory cannot be the same as the source project directory.\n"
+                    "Please choose a different folder for the packaged project."
+                ),
+            )
+            self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(
+                False
+            )
```
suricactus: If I read it right, we already made the button inactive on line 211, so do we need this line?

qfieldsync/gui/package_dialog.py
```diff
@@ -211,6 +211,28 @@ def package_project(self):
...
+            return
+        else:
+            self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(
+                True
+            )
```
suricactus:
```suggestion

        self.button_box.button(QDialogButtonBox.StandardButton.Save).setEnabled(True)
```

Since we have early return above, we can skip the else part.

suricactus (changes requested):

qfieldsync/gui/package_dialog.py
SeqLaz: Yes, you are completely right!.....  It is not needed there!

qfieldsync/gui/package_dialog.py
SeqLaz: Indeed!

PR #732, opened 2025-10-23
feat: make third-party login cancellable and smoother

gounux:
This PR intends to make the third-party login behave more smoothly in case the user closes the browser tab open by the QGIS OAuth2 config. Also, it makes the third-party sign in cancellable with a button.

| 1. First opening of Sign In Dialog | 2. Clicking on `Sign In with XXX` | 3. After having clicked on `Cancel Sign In` (or timeout expired) |
| --- | --- | --- |
| image | image | image |

Also, the duration for going through 3rd party ID provider sign in is set to 60 seconds, a bit more than the previous 30 seconds ; the idea being to give more time to complete the sign in process.

### DIFF
--- a/qfieldsync/ui/cloud_login_dialog.ui
+++ b/qfieldsync/ui/cloud_login_dialog.ui
@@ -189,8 +189,8 @@
        <widget class="QLabel" name="ssoSpinnerLabel">
         <property name="maximumSize">
          <size>
-          <width>36</width>
-          <height>36</height>
+          <width>24</width>
+          <height>24</height>
          </size>
         </property>
         <property name="text">


gounux (commented):
@lukasgraf @nirvn this is the PR about what we talked earlier, thanks a lot for the idea !

In the current state of the PR, there is a segfault happening when closing QGIS afterwards. Is this maybe related to the event loop ? i.e. could we play a bit more with signals to avoid this ?  
Did not really investigate this direction, maybe it could be worth giving a try,

qfieldsync/ui/cloud_login_dialog.ui
```diff
@@ -179,6 +179,41 @@
...
+     <property name="title">
+      <string/>
+     </property>
+     <layout class="QFormLayout" name="formLayout_2">
+      <item row="0" column="0">
+       <widget class="QLabel" name="ssoSpinnerLabel">
+        <property name="maximumSize">
+         <size>
+          <width>36</width>
+          <height>36</height>
```
nirvn: This looks quite big on the screenshot, I'd match the height of the spinner with that of the cancel sign in button.

nirvn (commented):
Other than the spinner size, the code looks OK. I'd like to give a try and reshuffling the GUI a bit on my end but I'd wait until the PR is merged.

PR #734, opened 2025-11-23
Fix autopush to QFieldCloud setting gone missing

nirvn:
While showcasing QField[Cloud] during FOSS4G 2025, I noticed our autopush ot QFieldCloud spinbox had gone missing from the interface :scream: . The PR fixes this regression. Proof of resurrection: 

image


gounux (approved):
Thanks for bringing this autopush back to life again !

PR #738, opened 2026-01-11
Fix a couple of Qt6 enum errors

nirvn:
This fixes a serious issues when running QFieldSync with a Qt6 QGIS build, namely uploading of cloud project fails.

It also deals with an irritating but harmless error when logging into QFC.


gounux (approved):

nirvn:
@gounux , thanks, let's rebase your other PR to see if I missed anything.

PR #745, opened 2026-01-30
Fix error handling presence of gpkg wal/shm files on the cloud

nirvn:
@beanzmo stumbled on a silent killer in our QFieldSync code! :wink: 

The code handling the presence of wal and shm files within a cloud project remote files list was broken. The code referred to an nonexistent self.local_path variable, which crashed the copying process.

If you were lucky and the wal file was the first file, you'd end up with an empty directory indicating something very wrong, But, if it happened halfway, you would end up with a half copied cloud project. This could lead to failure to copy attachments, offline data data.gpkg, etc.

### DIFF
--- a/qfieldsync/core/cloud_transferrer.py
+++ b/qfieldsync/core/cloud_transferrer.py
@@ -400,15 +400,17 @@ class CloudTransferrer(QObject):
             if not dest_path.parent.exists():
                 dest_path.parent.mkdir(parents=True)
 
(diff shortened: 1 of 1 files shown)


suricactus (approved):
Good to merge.

Since you had to decrypt this now, I would suggest to add a few comments on the lines before merging, e.g. why do we continue if -shm or -wal files.

suricactus:
I think we better release this sooner than later, I can live without the comments for now.

PR #750, opened 2026-03-20
fix: add trailing slash to cloud API URL if missing

Rakanhf:
When logging in in QFieldsync to a custom URL and the URL doesn't end with a `/` it effects buttons like `Edit on QFieldCloud` and `Forgot password`


duke-nyuki:
Task linked: [QF-7928 QFieldSync: Edit on qfieldcloud link dependent on URL ending with `/`](https://app.clickup.com/t/2192114/QF-7928)

suricactus (approved):
Nice addition!

PR #754, opened 2026-04-08
fix(login): Align SSO login layout to web

toebivankenoebi:
Align Login dialog layout when SSO is enabled to the layout in QField and QFieldCloud.

- Before
image

- After
image

Login flow successfully tested with Entra ID.

### DIFF
--- a/qfieldsync/gui/cloud_login_dialog.py
+++ b/qfieldsync/gui/cloud_login_dialog.py
@@ -36,8 +36,6 @@ from qgis.PyQt.QtWidgets import (
     QGroupBox,
     QMainWindow,
     QPushButton,
-    QSizePolicy,
-    QSpacerItem,
     QWidget,
 )
 from qgis.PyQt.uic import loadUiType
@@ -197,8 +195,8 @@ class CloudLoginDialog(QDialog, CloudLoginDialogUi):
 
     def clear_login_widgets(self) -> None:
         self.set_login_groupbox_visibility(self.signInUsernameGroupBox, False)
+        self.authenticationDivider.setVisible(False)
 
-        # clear sso login buttons
         for push_button in self._sso_login_buttons:
             push_button.deleteLater()
 
@@ -245,22 +243,20 @@ class CloudLoginDialog(QDialog, CloudLoginDialogUi):
 
         self.clear_login_widgets()
 
-        # add vertical space before SSO login buttons
-        vertical_spacer = QSpacerItem(
-            0, 10, QSizePolicy.Policy.Minimum, QSizePolicy.Policy.Fixed
-        )
-        self.signInUsernameGroupBox.layout().addItem(vertical_spacer)
+        has_credentials = False
 
         for auth_method in auth_methods:
-            # credentials login: enabled static groupbox.
             if auth_method["id"] == "credentials":
                 self.set_login_groupbox_visibility(self.signInUsernameGroupBox, True)
+                has_credentials = True
                 continue
 
-            # sso provider: dynamically generate button to login.
-            login_button = QPushButton(
-                self.tr("Sign In with {provider}").format(provider=auth_method["name"])
-            )
+            # Mirror the web login page: show an "Or" divider between the credentials
+            # form and the SSO provider buttons only when both are present.
+            if has_credentials and not self._sso_login_buttons:
+                self.authenticationDivider.setVisible(True)
+
+            login_button = QPushButton(auth_method["name"])
             login_button.setCursor(QCursor(Qt.CursorShape.PointingHandCursor))
 
             self.set_sso_provider_button_style(auth_method.get("styles"), login_button)
@@ -268,7 +264,7 @@ class CloudLoginDialog(QDialog, CloudLoginDialogUi):
             login_button.clicked.connect(
                 partial(self.on_login_with_sso_provider_button_clicked, auth_method)
             )
-            self.signInUsernameGroupBox.layout().addWidget(login_button)
+            self.signInUsernameGroupBox.layout().addRow(login_button)
             self._sso_login_buttons.append(login_button)
 
     def set_sso_provider_button_style(
--- a/qfieldsync/ui/cloud_login_dialog.ui
+++ b/qfieldsync/ui/cloud_login_dialog.ui
@@ -144,6 +144,72 @@
         </property>
        </widget>
       </item>
+      <item row="4" column="0" colspan="2">
+       <widget class="QWidget" name="authenticationDivider" native="true">
+        <property name="visible">
+         <bool>false</bool>
+        </property>
+        <layout class="QHBoxLayout" name="authenticationDividerLayout">
+         <property name="leftMargin">
+          <number>0</number>
+         </property>
+         <property name="topMargin">
+          <number>4</number>
+         </property>
+         <property name="rightMargin">
+          <number>0</number>
+         </property>
+         <property name="bottomMargin">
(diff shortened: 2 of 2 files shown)


nirvn:
@toebivankenoebi , nice one, much cleaner.

toebivankenoebi:
Force pushed because of linting error fixes.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -245,21 +247,48 @@ def on_fetch_auth_methods_finished(self) -> None:
...
+                divider_layout.addWidget(left_line)
+                divider_layout.addWidget(or_label)
+                divider_layout.addWidget(right_line)
+
+                self.signInUsernameGroupBox.layout().addRow(divider)
+                self._sso_login_buttons.append(divider)
+
             login_button = QPushButton(
-                self.tr("Sign In with {provider}").format(provider=auth_method["name"])
+                self.tr("{provider}").format(provider=auth_method["name"])
```
suricactus:
```suggestion
                self.tr(provider).format(provider=auth_method["name"])
```

No need to translate a dynamic string :)

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -245,21 +247,48 @@ def on_fetch_auth_methods_finished(self) -> None:
...
                 self.set_login_groupbox_visibility(self.signInUsernameGroupBox, True)
+                has_credentials = True
                 continue
 
-            # sso provider: dynamically generate button to login.
+            # Mirror the web login page: show an "Or" divider between the credentials
+            # form and the SSO provider buttons only when both are present.
+            if has_credentials and not self._sso_login_buttons:
+                divider = QWidget()
+                divider_layout = QHBoxLayout(divider)
```
suricactus: Wouldn't it be easier if we define all of these statically in the `.ui` file and show-hide them when needed, instead of building the whole thing from scratch in Python?

suricactus (commented):
Looks much nicer, thanks for that (as well as the `bash` script ;) ).

Few suggestions to reduce code.

qfieldsync/gui/cloud_login_dialog.py
m-kuhn:
```suggestion
                auth_method["name"]
```

no need to translate a dynamic string, take II

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -288,7 +288,7 @@ def on_fetch_auth_methods_finished(self) -> None:
                 self._sso_login_buttons.append(divider)
 
             login_button = QPushButton(
-                self.tr("{provider}").format(provider=auth_method["name"])
+                self.tr(provider).format(provider=auth_method["name"])
```
m-kuhn: I think this does only half the job, see https://github.com/opengisch/qfieldsync/pull/754#discussion_r3051263039

m-kuhn: Actually, I think this breaks things, as there is no `{provider}` placeholder to replace for the `format` function

toebivankenoebi: I'll change/improve/check and come back.

qfieldsync/gui/cloud_login_dialog.py
toebivankenoebi: Yes, will put the divider into the ui and set visibility = true if sso.

qfieldsync/gui/cloud_login_dialog.py
suricactus: sorry, too focused on the dynamic part to see the rest of the code :)

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -245,30 +243,28 @@ def on_fetch_auth_methods_finished(self) -> None:
...
                 continue
 
-            # sso provider: dynamically generate button to login.
-            login_button = QPushButton(
-                self.tr("Sign In with {provider}").format(provider=auth_method["name"])
-            )
+            # Mirror the web login page: show an "Or" divider between the credentials
+            # form and the SSO provider buttons only when both are present.
+            if has_credentials and not self._sso_login_buttons:
+                self.orDivider.setVisible(True)
```
suricactus:
```suggestion
                self.authenticationDivider.setVisible(True)
```
I believe we can have slightly better name here if you don't mind :)

Not so strong opinion about the exact name, just avoid starting with `or` or `and` as a name, looks like someone lost the first half.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -245,30 +243,28 @@ def on_fetch_auth_methods_finished(self) -> None:
...
+
+            login_button = QPushButton(self.tr(auth_method["name"]))
```
suricactus:
```suggestion
            login_button = QPushButton(auth_method["name"])
```

Same as before, no need to translate a dynamic string.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -165,7 +163,7 @@ def __init__(
         self.rejected.connect(self.on_rejected)
         self.hide()
 
-        self._sso_login_buttons: List[QPushButton] = []
+        self._sso_login_buttons: List[QWidget] = []
```
suricactus:
```suggestion
        self._sso_login_buttons: List[QPushButton] = []
```

I think this change is no longer needed, the `login_buttons` are `QPushButton`, which is a `QWidget` instance as a parent.

qfieldsync/gui/cloud_login_dialog.py
```diff
@@ -197,10 +195,10 @@ def toggle_server_url_visibility(self) -> None:
 
     def clear_login_widgets(self) -> None:
         self.set_login_groupbox_visibility(self.signInUsernameGroupBox, False)
+        self.orDivider.setVisible(False)
 
-        # clear sso login buttons
-        for push_button in self._sso_login_buttons:
-            push_button.deleteLater()
+        for button in self._sso_login_buttons:
+            button.deleteLater()
```
suricactus:
```suggestion
        for push_button in self._sso_login_buttons:
            push_button.deleteLater()
```

`_sso_login_buttons` is still a `QPushButton`. The renaming is unnecessary.

qfieldsync/gui/cloud_login_dialog.py
toebivankenoebi: makes sense, done

qfieldsync/gui/cloud_login_dialog.py
toebivankenoebi: Thanks, Done.

qfieldsync/gui/cloud_login_dialog.py
toebivankenoebi: Of course, thanks, done.

qfieldsync/gui/cloud_login_dialog.py
toebivankenoebi: done.

toebivankenoebi:
@suricactus : All comments addressed, some tests are failing even though test cases were successful. If that's a known issue and you're happy with the changes I will squash, clean up commit history and force push.

suricactus:
Should have been fixed in #755 . Please rebase and rephrase/squash commit messages

suricactus (approved):

PR #756, opened 2026-04-09
fix: stream large files using QFile.setBodyDevice to avoid 2GB limit

### LINKED ISSUE #751: Upload of Datasets >2GB via QFieldSync fails successfully

Hello,

uploading larger files (in my case: .tiff) via QFieldSync into QFieldCloud fails - QFieldSync says the file upload was successfull though.

Image

 In the Cloud, the file is there - but with only 3 Bytes.

Image

Upload via SDK worked surprisingly - so the issue is purely QFieldSync related.

1. Upload raster >2GB into QFieldCloud using QFieldSync
2. QFieldSync reports successfull transfer into the cloud
3. Checking the raster file via app.qfield.cloud -> file size shows 3 Byte

### QGIS version

3.44

### QFieldSync Version

4.18.3

### Operating system name

Windows

### Operating system version

Windows 11

### Reinstall QFieldSync

- [x] I have a fresh install of the latest QFieldSync version on a new QGIS profile, but the problem persists.
- [x] Problem happens with all files and projects, not only some files or projects.

### Additional context

_No response_

meyerlor:
QHttpPart.setBody() loads the entire file into a QByteArray, which uses a 32-bit signed integer for size. Files >= 2GB overflow this limit, causing the upload to silently succeed while only writing ~3 bytes to the server.

Fix by using setBodyDevice(QFile) instead, which streams the file directly without loading it into memory. QFile is parented to multi_part (which is parented to reply) to ensure it stays open for the full duration of the request.

Compared it to qfieldcloud-sdk solution and found+created the fix with the help of claude.ai !

Built a local release and tested successfully with QGIS 3.44 on Ubuntu and a 2.5GB .tif.

grafik

grafik

Fixes #751

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -39,6 +39,7 @@ from qgis.core import (
 from qgis.PyQt.QtCore import (
     QByteArray,
     QEventLoop,
+    QFile,
     QFileSystemWatcher,
     QObject,
     QUrl,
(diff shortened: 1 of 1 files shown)


suricactus:
@meyerlor thanks for the contribution. Looks promising. Would you mind merging/rebasing your PR on the current master, which fixed the CI.

meyerlor:
@suricactus happy to help! I force pushed / rebased!

suricactus:
Thanks @meyerlor !

PR #766, opened 2026-04-22
Add cookie saving/restoring to avoid need for SSO login for each QGIS session

nirvn:
@suricactus , @lukasgraf , as discussed over Google Chat.

@lukasgraf , could you give this a try with your local instance?

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -38,6 +38,7 @@ from qgis.core import (
 )
 from qgis.PyQt.QtCore import (
     QByteArray,
+    QDateTime,
     QEventLoop,
     QFile,
     QFileSystemWatcher,
@@ -49,6 +50,7 @@ from qgis.PyQt.QtCore import (
 from qgis.PyQt.QtNetwork import (
     QHttpMultiPart,
     QHttpPart,
+    QNetworkCookie,
     QNetworkReply,
     QNetworkRequest,
 )
@@ -248,6 +250,37 @@ class CloudNetworkAccessManager(QObject):
 
         # use the default URL
         self.set_url(url)
+
+        self.restore_cookies()
+
+    def restore_cookies(self):
+        raw_cookies = self.preferences.value("qfieldCloudCookies")
+        for raw_cookie in raw_cookies:
+            cookies = QNetworkCookie.parseCookies(raw_cookie)
+            if (
+                cookies
+                and QDateTime.currentSecsSinceEpoch()
+                < cookies[0].expirationDate().toSecsSinceEpoch()
+            ):
+                self._nam.cookieJar().insertCookie(cookies[0])
+
+    def save_cookies(self):
+        raw_cookies = []
+        cookies = self._nam.cookieJar().cookiesForUrl(QUrl(self.url))
+        for cookie in cookies:
+            if (
+                not cookie.isSessionCookie()
+                and QDateTime.currentSecsSinceEpoch()
+                < cookie.expirationDate().toSecsSinceEpoch()
+            ):
+                raw_cookies.append(cookie.toRawForm())
+
+        self.preferences.set_value("qfieldCloudCookies", raw_cookies)
+
+    def delete_cookies(self):
+        cookies = self._nam.cookieJar().cookiesForUrl(QUrl(self.url))
+        for cookie in cookies:
+            self._nam.cookieJar().deleteCookie(cookie)
 
     def handle_response(
         self, reply: QNetworkReply, should_parse_json: bool = True
@@ -953,6 +986,8 @@ class CloudNetworkAccessManager(QObject):
             self.preferences.set_value("qfieldCloudAuthcfg", "")
             self.auth_config = None
             self.preferences.set_value("qfieldCloudAuthMethod", self.auth_method.value)
+            self.delete_cookies()
+            self.save_cookies()
             self.logout_success.emit()
         except QfcError as err:
             self.logout_failed.emit(str(err))
@@ -962,6 +997,7 @@ class CloudNetworkAccessManager(QObject):
         try:
             self.json_object(reply)
             self.clear_sso_config()
+            self.delete_cookies()
             self.logout_success.emit()
         except QfcError as err:
             self.logout_failed.emit(str(err))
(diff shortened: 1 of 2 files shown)


suricactus (commented):
Code looks good, I leave @lukasgraf for the final review.

qfieldsync/core/cloud_api.py
```diff
@@ -248,6 +250,37 @@ def __init__(self, parent=None) -> None:
         # use the default URL
         self.set_url(url)
 
+        self.restore_cookies()
+
+    def restore_cookies(self):
+        raw_cookies = self.preferences.value("qfieldCloudCookies")
```
lukasgraf, line 256: So, after switching the previous version of my qfieldsync plugin to this branch, I get the following traceback on QGIS launch:

```pytb
NameError: QFieldSync has no setting qfieldCloudCookies 

Traceback (most recent call last):
  File "/Applications/QGIS-final-3_44_4.app/Contents/Frameworks/lib/python3.12/site-packages/qgis/utils.py", line 478, in _startPlugin
    plugins[packageName] = package.classFactory(iface)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasgraf/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/__init__.py", line 58, in classFactory
    return QFieldSync(iface)
           ^^^^^^^^^^^^^^^^^
  File "/Users/lukasgraf/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/qfield_sync.py", line 132, in __init__
    self.network_manager = CloudNetworkAccessManager(self.iface.mainWindow())
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasgraf/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/core/cloud_api.py", line 253, in __init__
    self.restore_cookies()
  File "/Users/lukasgraf/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/core/cloud_api.py", line 256, in restore_cookies
    raw_cookies = self.preferences.value("qfieldCloudCookies")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lukasgraf/Library/Application Support/QGIS/QGIS3/profiles/default/python/plugins/qfieldsync/setting_manager/setting_manager.py", line 53, in value
    raise NameError('%s has no setting %s' % (self.plugin_name, setting_name))
NameError: QFieldSync has no setting qfieldCloudCookies
```

Is there possibly some sort of preferences migration that the plugin would need to perform, or do we need to make the `self.preferences.value("qfieldCloudCookies")` access more defensive, dealing with the fact that that preference key might not exist?

It looks like the new preferences key isn't even added in `preferences.py`?

lukasgraf (changes requested):

qfieldsync/core/cloud_api.py
nirvn: Oh yeah I forgot to commit that file.

nirvn: It's pretty straightforward though. We add a global scope string list preference with an empty array as default value

lukasgraf: So even when I make that change manually, I can't currently get a successful signin.
I get varying errors displayed in red in the QFS UI, like `Expecting value: line 1 column 1 (char 0)`, or sometimes a truncated dump of HTML, with no immediately obvious way to tell what's happening.

In this version, with the current state in my profile, QFS is not allowing me to really test this change. I need to focus on some other stuff first, and come back later to review this.

nirvn: @lukasgraf , no worries, I'll reach out when that's fixed

nirvn: @lukasgraf , missing bit added.

suricactus: @lukasgraf ping

lukasgraf (changes requested):
Unfortunately, this doesn't work as expected for me.

The cookies that are getting saved and restored are always empty (`[]`). They are present in the cookie jar at the beginning of [`_on_get_user_info_finished()`](https://github.com/opengisch/qfieldsync/blob/sso_cookies/qfieldsync/core/cloud_api.py#L1021), but they get emptied as soon as the [`self.get_file()` for the avatar](https://github.com/opengisch/qfieldsync/blob/sso_cookies/qfieldsync/core/cloud_api.py#L1040-L1043) has finished executing.

The reason for that seems to be [this call to `self._clear_cloud_cookies()`in `cloud_get()`](https://github.com/opengisch/qfieldsync/blob/sso_cookies/qfieldsync/core/cloud_api.py#L685).

nirvn:
@lukasgraf , OK, another round. While I don't have the proper setup locally to test things, I _am_ seeing non-SSO cookie saved and restored across sessions. So my level of confidence is increasing ;)

suricactus:
@nirvn what is the cookie name?

nirvn:
@suricactus , the one I see is sessionid

lukasgraf:
@nirvn I gave this another test, but for me it doesn't seem to keep the Django session to QFC across QGIS restarts. What I observe is that QFS attempts to trigger the OAuth flow on every restart of QGIS. Even if one before was successful, and it therefore should have a Django session that is still valid.

nirvn:
@lukasgraf , are those two distinct issues though? Cookie remembrance across session vs. QFS / QGIS' OAuth2 provider behavior?

lukasgraf:
No. Successful OAuth2 authentication in QFC _will_ return a response with the cookies that establish a Django session.

nirvn:
@lukasgraf , and those cookies are not sent by QFC? Or are the cookies sent but nevertheless still not remembered across sessions?

lukasgraf:
They're sent, but not remembered.
I will have to dig into the code again with a debugger to tell exactly what's happening.

lukasgraf (approved):
As tested together, finally works! 👍 🎉

PR #770, opened 2026-04-24
Implement a QFC storage widget and add it in the project creation and synchronization dialogs

nirvn:
This PR implements a storage widget to display a given QFC user's used and available storage. As with QField, if the user is getting close or as used 100% of their available storage, they will be invited to upgrade via an hyperlink attached to the storage widget.

The storage widget is being used in two places:
- Within the project creation dialog; and
- Within the project synchronization dialog

Screenshot time.

image1

image2

### DIFF
--- a/qfieldsync/gui/storage_widget.py
+++ b/qfieldsync/gui/storage_widget.py
@@ -1,0 +1,162 @@
+"""
+/***************************************************************************
+ StorageWidget
+                                A QGIS plugin to
+                           Sync your projects to QField
+                             -------------------
+        begin                : 2026-04-24
+        git sha              : $Format:%H$
+        copyright            : (C) 2026 by OPENGIS.ch
+        email                : info@opengis.ch
+ ***************************************************************************/
+
+/***************************************************************************
+ *                                                                         *
+ *   This program is free software; you can redistribute it and/or modify  *
+ *   it under the terms of the GNU General Public License as published by  *
+ *   the Free Software Foundation; either version 2 of the License, or     *
+ *   (at your option) any later version.                                   *
+ *                                                                         *
+ ***************************************************************************/
+"""
+
+import os
+
+from qgis.PyQt.QtWidgets import QWidget
+from qgis.PyQt.uic import loadUiType
+
+from qfieldsync.core.cloud_api import CloudNetworkAccessManager, QfcError
+from qfieldsync.utils.file_utils import filesizeformat10
+
+StorageWidgetUi, _ = loadUiType(
+    os.path.join(os.path.dirname(__file__), "../ui/storage_widget.ui")
+)
+
+
+class StorageWidget(QWidget, StorageWidgetUi):
+    WARNING_THRESHOLD = 0.8
+    CRITICAL_THRESHOLD = 0.95
+
+    NORMAL_COLOR = "#4a6fae"
+    WARNING_COLOR = "#ffa500"
+    CRITICAL_COLOR = "#c0392b"
+
+    PROGRESSBAR_STYLESHEET = "QProgressBar {{ border: 1px solid palette(Mid); border-radius: 2px; }} QProgressBar::chunk {{ background-color: {}; }}"
+
+    def __init__(
+        self, network_manager: CloudNetworkAccessManager, parent: QWidget = None
+    ) -> None:
+        super().__init__(parent=parent)
+        self.setupUi(self)
+
+        self.progressBar.setMaximum(100)
+        self.progressBar.setStyleSheet(
+            self.PROGRESSBAR_STYLESHEET.format(self.NORMAL_COLOR)
+        )
+
+        self.network_manager = network_manager
+
+        self._owner = ""
+        self._active_storage_total_bytes = 0
+        self._storage_used_bytes = 0
+        self._storage_used_percent = 0.0
+        self._storage_management_hyperlink = ""
+
+    def set_owner(self, owner: str) -> None:
+        self._owner = owner
+        if self._owner:
+            self.refresh()
+
+    def owner(self) -> str:
+        return self._owner
+
+    def active_storage_total_bytes(self) -> int:
+        return self._active_storage_total_bytes
+
+    def storage_used_bytes(self) -> int:
+        return self._storage_used_bytes
+
+    def storage_used_percent(self) -> float:
+        return self._storage_used_percent
+
+    def storage_management_hyperlink(self) -> str:
+        return self._storage_management_hyperlink
+
+    def refresh(self) -> None:
+        if self._owner:
+            self.progressBar.setMaximum(0)
+            self.usageLabel.setText(self.tr("Fetching storage details..."))
+
+            reply = self.network_manager.get_subscription_information(self._owner)
+            reply.finished.connect(
+                lambda: self.on_get_subscription_information_finished(reply)
+            )
+
+    def on_get_subscription_information_finished(self, reply) -> None:
+        try:
+            subscription_information = self.network_manager.json_object(reply)
+        except QfcError:
+            subscription_information = {}
+
+        self.progressBar.setMaximum(100)
+
+        reply.deleteLater()
+
+        self._active_storage_total_bytes = subscription_information.get(
+            "active_storage_total_bytes", 0
+        )
+        self._storage_used_bytes = subscription_information.get("storage_used_bytes", 0)
+
+        if self._active_storage_total_bytes > 0:
+            self._storage_used_percent = (
+                self._storage_used_bytes / self._active_storage_total_bytes
+            )
+        else:
+            self._storage_used_percent = 0
+
+        self.progressBar.setValue(int(self._storage_used_percent * 100))
+        self.progressBar.setStyleSheet(
+            self.PROGRESSBAR_STYLESHEET.format(
+                self._get_progressbar_color(self._storage_used_percent)
+            )
+        )
+
+        usage_label = self.tr("Used {} of {}").format(
+            filesizeformat10(self._storage_used_bytes),
+            filesizeformat10(self._active_storage_total_bytes),
+        )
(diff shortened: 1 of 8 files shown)


suricactus:
I am not convinced by the progress bar here, because in desktop applications it immediately triggers the association of, well, process progress. We either need to style the progress bar in a way not to be confused with a regular progress bar, or we can just hide it for now.

nirvn:
@suricactus , +1 to style. Good idea

nirvn:
@suricactus , style updated, you can see how it looks via the updated screenshot in the description.

suricactus:
Lovely and in sync with her rest of the UI.

My last question is shall we try to put it on top as we do in QField?

This will give a bit of coherence across the apps. On QFC is already more on the top rather than bottom?

nirvn:
@suricactus , I had it on top before, and TBH it did not look good in QFS. 

First, for the project creation, it just looks a lot better tied to the owner combobox. For the cloud file transfer dialog, it also felt off, even when I added a title label on top.

If we wanted it on top, we'd probably have to invest in UI redesign a bit more than we'd consider worth it :)

MidnightCoder-m:
Lovely!

nirvn:
Out of storage check + UI feedback when user about to upload file(s) exceeding remaining available storage.

When the user is able to manage/upgrade, we invite the user to do so:

Screenshot From 2026-04-26 19-23-59

Otherwise, we invite the user to try and reduce the total file size being uploaded:

Screenshot From 2026-04-26 19-24-29

qfieldsync/gui/storage_widget.py
```diff
@@ -0,0 +1,188 @@
...
+        else:
+            self._active_storage_total_bytes = 0
+            self._storage_used_bytes = 0
+            self._storage_used_percent = 0.0
+
+            self.progressBar.setValue(0)
+            self.progressBar.setStyleSheet(
+                self.PROGRESSBAR_STYLESHEET.format(self.NORMAL_COLOR)
+            )
+            self.usageLabel.setText(self.tr("N/A"))
```
suricactus, line 142:
```suggestion
        try:
            subscription_information = self.network_manager.json_object(reply)
        except QfcError:
            subscription_information = {}

        self.progressBar.setMaximum(100)

        reply.deleteLater()

        self._active_storage_total_bytes = subscription_information.get(
            "active_storage_total_bytes"
        )
        self._storage_used_bytes = subscription_information.get(
            "storage_used_bytes", 0
        )
        
        if self._active_storage_total_bytes > 0:
            self._storage_used_percent = (
                self._storage_used_bytes / self._active_storage_total_bytes
            )
        else:
            self._storage_used_percent = 0

        self.progressBar.setValue(int(self._storage_used_percent * 100))
        self.progressBar.setStyleSheet(
            self.PROGRESSBAR_STYLESHEET.format(self._get_progressbar_color(self._storage_used_percent))
        )

        usage_label = self.tr("Used {} of {}").format(
            self.represent_file_size(self._storage_used_bytes),
            self.represent_file_size(self._active_storage_total_bytes),
        )
        if self._owner == self.network_manager.user_details.get("username"):
            usage_label += " / <a href='{}'>{}</a>".format(
                self._get_upgrade_link_url(subscription_information.get("is_premium"))
                self._get_upgrade_link_label(self._storage_used_percent)
            )
        else:
            self._storage_management_hyperlink = ""

        if subscription_information:
            self.usageLabel.setText(usage_label)
        else:
            self.usageLabel.setText(self.tr("N/A"))

    # I would even consider these two to go to `cloud_utils.py`
    def _get_progressbar_color(self, storage_used_percent: float) -> QColor:
        if storage_used_percent >= self.CRITICAL_THRESHOLD:
            return self.CRITICAL_COLOR
        elif storage_used_percent >= self.WARNING_THRESHOLD:
            return self.WARNING_COLOR
        else:
            return self.NORMAL_COLOR
            
     def _get_upgrade_link_label(self, storage_used_percent: float) -> str:
         if storage_used_percent >= self.CRITICAL_THRESHOLD:
            upgrade_label = self.tr("upgrade storage") 
        else:
            upgrade_label = self.tr("manage storage")
      
      def _get_upgrade_link_url(self, is_premium: bool) -> str:
          if is_premium:
              return f"https://app.qfield.cloud/settings/{self._owner}/billing"
          else:
              return "https://app.qfield.cloud/plans"
            
```

Instead of aditional branching, we can reduce two levels of `if` by simply using a `dict` instead of `None` as a default value of the `subscription_information` variable.

Also a few code snippets can be split into their own self-contained functions, so they don't bring additional noise in the already long function.

Finally, instead of using `community` plan name, you can use `is_premium` flag to determine what upgrade path to take.

qfieldsync/gui/storage_widget.py
```diff
@@ -0,0 +1,188 @@
...
+        return self._active_storage_total_bytes
+
+    def storage_used_bytes(self) -> int:
+        return self._storage_used_bytes
+
+    def storage_used_percent(self) -> float:
+        return self._storage_used_percent
+
+    def storage_management_hyperlink(self) -> str:
+        return self._storage_management_hyperlink
```
suricactus, line 83: I am not sure if we really need these as methods, when we have the properties, but let's keep it like this for now.

qfieldsync/gui/storage_widget.py
```diff
@@ -0,0 +1,188 @@
...
+        self.usageLabel.setFont(font)
+
+        self.progressBar.setSizePolicy(
+            QSizePolicy.Policy.MinimumExpanding, QSizePolicy.Policy.Maximum
+        )
+        self.progressBar.setTextVisible(False)
+        self.progressBar.setMinimumHeight(8)
+        self.progressBar.setMaximumHeight(8)
+        self.progressBar.setMinimum(0)
+        self.progressBar.setMaximum(100)
```
suricactus: What is the reason to defined each of these here, instead of the XML `.ui` file directly?

qfieldsync/gui/storage_widget.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def represent_file_size(self, file_size: int) -> str:
+        divider = 1000.0
+        for unit in ("", "KB", "MB", "GB"):
+            if abs(file_size) < divider:
+                return f"{file_size:3.1f}{unit}"
+
+            file_size /= divider
+
+        return f"{file_size:.1f}TB"
```
suricactus: I would suggest we use exactly the same function as we do in QFC: https://github.com/opengisch/QFieldCloud/blob/bf990a42098e8ad5f18047fa9f6fd9b97b0c6045/docker-app/qfieldcloud/core/templatetags/filters.py#L11-L53

It can happily live in `file_utils.py`.

suricactus (commented):
Thanks! The new progressbar looks much better than the initial version and much more in sync with QField (and even QFieldCloud).

> First, for the project creation, it just looks a lot better tied to the owner combobox. For the cloud file transfer dialog, it also felt off, even when I added a title label on top.
>
> If we wanted it on top, we'd probably have to invest in UI redesign a bit more than we'd consider worth it :)

I see, thank you! We need to rework the overall GUI in all three clients, which is out of scope of this PR.

---

You need to rebase on latest `master` and ensure you updated the local `pre-commit` config.

qfieldsync/gui/storage_widget.py
nirvn: @suricactus , I've always preferred doing this on the coding side of things.

qfieldsync/gui/storage_widget.py
nirvn: @suricactus , can we do that as a follow up set of libqfieldsync, qfieldsync, qfieldcloud PRs?

qfieldsync/gui/storage_widget.py
nirvn: @suricactus , applied.

nirvn:
@suricactus , thanks for the review. The big refactoring you suggested has been applied. I'd suggest we defer the file size string representation in a follow up set of PRs as it'll touch [lib]qfield[sync|cloud] :)

qfieldsync/gui/storage_widget.py
suricactus: But why? It makes it much easier to anyone to manipulate it in qt designer, now checking a styling of a text, e.g. the italic has to be done in two different places - qt designer and python code.

Unless a big -1, please move this logic to the qt's ui file, the majority if not all code here comes with that assumption.

qfieldsync/gui/storage_widget.py
suricactus: Why libqfieldsync? What about here: https://github.com/opengisch/qfieldsync/blob/master/qfieldsync/utils/file_utils.py

Copy-paste the implementation for now. You can reference the source on qfieldcloud.

qfieldsync/gui/storage_widget.py
nirvn: You evidentially feel strongly about it, more so than I do, can move.

I don't really use Qt Designer to begin with (I dislike the massive changes it can trigger in .ui when merely changing a property). That's probably why I prefer regrouping everything in Python beyond the generic widget tree.

nirvn:
@suricactus , extra comments addressed.

qfieldsync/utils/file_utils.py
```diff
@@ -211,7 +213,7 @@ def _open_with_onedrive_retry(
         filename: Path to the file to open.
         mode: File open mode (default ``"rb"``).
         max_retries: Number of retries after the first failure.
-        retry_delay_s: Base delay in seconds between retries (doubles each attempt).
+        retry_delay_s: Base delay in seconds between retries (doublfrom qgis.PyQt.QtCore import QObjectes each attempt).
```
suricactus: I don't really follow the change in the commen here?

suricactus (approved):
Nice! Last comment about a comment.

PR #771, opened 2026-04-26
ci: use qgis's `pyqgis4-checker` image

gounux:
The image for checking pyqgis4 / PyQt6 compatibility is now maintained via the QGIS org, see https://github.com/qgis/pyqgis4-checker

See https://github.com/opengisch/libqfieldsync/pull/145 that does pretty much the same for `libqfieldsync`.


suricactus:
Thanks!

suricactus (approved):

PR #777, opened 2026-04-27
Remove tmate session from the CI

suricactus:
It imposes risks and it is used only in the very rare cases needed for debugging why the CI fails.


m-kuhn (approved):
thanks

PR #778, opened 2026-04-27
Remove completely redundant scripts directory as nobody uses the scripts in there

suricactus:
We have pre-commit CLI to handle.


m-kuhn (approved):
thanks for the cleanup

PR #779, opened 2026-04-28
Let Qt handle styling for the read-only local project path line edit

nirvn:
I keep stumbling on this and forget to fix it. No more!

In the cloud transfer dialog, when using a dark theme, the local project line edit is essentially unreadable:

image1-3

The PR fixes the situation by removing the custom stylesheet we apply there and revert to Qt's default handling.


suricactus (approved):
Nice!

suricactus:
Before merging, are you sure you tested it on light theme too?

nirvn:
@suricactus , yep I did. Checked against fusion light / dark, night mapping, and blend of gray.

PR #780, opened 2026-04-28
Detect and remove dead code from the codebase

suricactus:
The main downside is we have to explicitly mark interface implementation with `# dead: disable` but this is rare in QFieldSync so we can afford it.


nirvn (approved):

nirvn:
Good one

PR #783, opened 2026-04-30
Use simple docker commands to run the tests

suricactus:
Same as we did on libqfieldsync. https://github.com/opengisch/libqfieldsync/pull/151/changes


nirvn (approved):
Big +1, thanks.

PR #792, opened 2026-05-10
Add a new `Config` class that sets project level configurations to the qgs file

suricactus:
Previously one was required to use `ProjectProperties` and to explicitly define a getter and setter for a given property. This was tiresome, repetitive
 and error prone.

Check libqfieldsync for full understanding of the change.

This is very similar to what `settings_manager` does and it will eventually replace it.

See https://github.com/opengisch/libqfieldsync/pull/156


duke-nyuki:
Task linked: [QF-8226 Switch to a easier to defined and use project settings manager in QFielddSync](https://app.clickup.com/t/2192114/QF-8226)

PR #795, opened 2026-05-12
fix: file storage widget fails to load due to Python error trying to call `tr()` statically on `QObject`

suricactus:
Just make a tmp `QObject`, used in several other places too.

Error:
```
TypeError: tr(self, sourceText: str, disambiguation: str = None, n: int = -1): first argument of unbound method must have type 'QObject'
```


duke-nyuki:
Task linked: [QF-8240 Fix TypeError: tr(self, sourceText: str, disambiguation: str = None, n: int = -1): first argument of unbound method must have type 'QObject'](https://app.clickup.com/t/2192114/QF-8240)

gounux (approved):
Checked out locally, problem solved thanks !

PR #799, opened 2026-05-14
Bump libqfieldsync fixing the copying files with their mode

### LINKED ISSUE #797: read-only geojson leads to crash on re-export

I have a project where one layer is a geojson which is mode 444.   I made it like that because I created it by hand; it's a boundary that was superceded in the1970s and my file is a synthetic arc between two current boundary points.   That took me some time and I don't want the file edited.  That is otherwise fine.

I have specified "Copy" for this layer.  I want to see it in the field (when visiting boundary monuments), but I don't want to change it.  This feels normal and is mostly fine.

Today I just did a bunch of editing in the project, dealing with  previous field  collection, and checked QField config, and then did "package for qfield".   I got a stack backtrace and it failed. The problem was permission denied copying the read-only layer in place.

The fix is to write to a temporary file and rename, probably.   Or to remove files before writing.   I'm really not sure, but this should work.

If you address this  please be extra careful not to delete anything outside the QField export dir.   Removing the whole dir feels way too dangerous, and just the files that are being written seems just barely safe.

Create a project with  mode 444 geojson layer.  Export it.  Export it again.

### QGIS version

3.44.9

### QFieldSync Version

4.21.1

### Operating system name

Other (please specify in additional context)

### Operating system version

NetBSD 10 amd64

### Reinstall QFieldSync

- [ ] I have a fresh install of the latest QFieldSync version on a new QGIS profile, but the problem persists.
- [ ] Problem happens with all files and projects, not only some files or projects.

### Additional context

I didn't do a reinstall but my plugin is up to date.   The backtrace is clear and once I read it this is not surprising.  Let me know if you find this puzzling.

suricactus:
See https://github.com/opengisch/libqfieldsync/pull/158
Fixes https://github.com/opengisch/qfieldsync/issues/797

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 # NOTE `libqfielsync` version should be defined in the `*.tar.gz` format, not `git+https://` to make `wheel` happy
-libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/178fcc14fa4a3898b9ad2c01e620013e654a1af5.tar.gz
+libqfieldsync @ https://github.com/opengisch/libqfieldsync/archive/b3541aa99630f39354e479c75f0f1bcb2f7020ba.tar.gz


duke-nyuki:
Task linked: [QF-8254 Fix read-only files leading to uncaught exceptions](https://app.clickup.com/t/2192114/QF-8254)

nirvn (approved):

PR #800, opened 2026-05-16
Add ability to make a cloud project public during the creation process

nirvn:
Title says it all. Screenshot:

image

Since we can control whether a cloud project is public or not during creation through app.qfield.cloud, and we now have a clear UI when editing project properties in QFS, it feels like we'd benefit from matching that UI and behavior for projection creation in QFS.


suricactus (approved):
Thanks!

PR #801, opened 2026-05-16
Present project properties' files details using friendly file sizes

nirvn:
Before (left) vs PR (right):

rect1

Tiny paper cut I stumbled upon when doing the storage meter last week. I can't unsee so better fix it :)


suricactus (approved):
Happy we have this function here.

nirvn:
@suricactus , yeah, somebody was right in asking me to make it a function hey? :wink:  good call.

PR #809, opened 2026-06-09
fix: enable login when there is at least one auth method available

### LINKED ISSUE #807: Not possible to authenticate towards qfieldcloud instance without password login (only SSO) (solution supplied)

The login form (username, password and SSO buttons) are disabled and never unlocked if SSO authentication is the only option in the qfieldcloud instance you are connecting to.

Try to authenticate towards only SSO enabled qfieldcloud instance

### QGIS version

3.44.8

### QFieldSync Version

4.21.3

### Operating system name

Windows

### Operating system version

Windows 11, Windows 10, Fedora 44

### Reinstall QFieldSync

- [x] I have a fresh install of the latest QFieldSync version on a new QGIS profile, but the problem persists.
- [x] Problem happens with all files and projects, not only some files or projects.

### Additional context

I tried to create a pull request but noticed that they where not open to everyone, which is understandable. But, it all comes down to the check that happens in gui/cloud_login_dialog.py at row 252.
```python
for auth_method in auth_methods:
    if auth_method["id"] == "credentials":
        self.set_login_groupbox_visibility(self.signInUsernameGroupBox, True)
        has_credentials = True
        continue
``` 
This means that the signInUsernameGroupBox never gets set to enabled. My solution whas to remove that line and instead, before the loop just set this instead:
```python
self.set_login_groupbox_visibility(
    self.signInUsernameGroupBox, len(auth_methods) > 0
)
``` 
I tested this and all tests pass and as far as I can see, works as intended. This is the test I added myself but maybe there more stuff that should be tested there.

[test_cloud_login_dialog.py](https://github.com/user-attachments/files/28719635/test_cloud_login_dialog.py)

gounux:
This PR intends to make QFieldSync cloud login dialog follow the targetted QFieldCloud instance configuration, e.g. login enabled or not the credentials login inputs depending on if the password login is enabled or not on the QFC instance.

If a QFieldCloud instance is configured with no password login, currently the login groupbox stays disabled, so it prevents the user from clicking on a 3rd party IDP login button.

Now, with this PR :

| Login **enabled** in QFieldCloud | Login **disabled** in QFieldCloud |
| --- | --- |
| image | image |
| Username & password line edits + `Sign In` button are **enabled**. | Username & password line edits + `Sign In` button are **disabled**. |

Also, getting rid of this `set_login_groupbox_visibility` method which was uncorrectly named and was only setting a groupbox enabled or not.

Fixes #807

### DIFF
--- a/qfieldsync/gui/cloud_login_dialog.py
+++ b/qfieldsync/gui/cloud_login_dialog.py
@@ -34,7 +34,6 @@ from qgis.PyQt.QtWidgets import (
     QApplication,
     QDialog,
     QDialogButtonBox,
-    QGroupBox,
     QMainWindow,
     QPushButton,
     QWidget,
@@ -132,7 +131,7 @@ class CloudLoginDialog(QDialog, CloudLoginDialogUi):
         )
 
         self.loginFormGroup.setVisible(False)
-        self.set_login_groupbox_visibility(self.signInUsernameGroupBox, False)
+        self.signInUsernameGroupBox.setEnabled(False)
 
         for server_url in self.network_manager.server_urls():
             self.serverUrlCmb.addItem(server_url)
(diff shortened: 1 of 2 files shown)


lukasgraf (approved):
Nice! LGTM! 👍 🎉

PR #811, opened 2026-06-18
Only throw an out of storage block when we are trying to upload content

nirvn:
Improve logic to insure we only throw an out of storage block when we are trying to upload content.


suricactus (approved):
Smart solution!

PR #818, opened 2026-07-16
Ensure the leading slashes are removed when dealing with QFieldCloud file requests

SeqLaz:


duke-nyuki:
Task linked: [QF-5477 Ensure the leading slashes '/' are removed when dealing with QFieldCloud file requests](https://app.clickup.com/t/2192114/QF-5477)

qfieldsync/core/cloud_api.py
```diff
@@ -936,10 +936,38 @@ def cloud_upload_files(
 
         return reply
 
+    def _validate_uri_path(self, uri: Union[str, list[str]]) -> None:
+        """Validate that incoming URI components do not contain dangerous path elements."""
+        forbidden_prefixes = ("/", "../", "./", "~/")
+
+        if isinstance(uri, str):
+            if "//" in uri:
+                raise ValueError("Bad Request: URL contains duplicate slashes '//'")
```
gounux: What if the URI contains the `://` scheme separator, e.g. `http://` or `https://` ?

Checked out and grepped locally, it seems that the `_prepare_uri` calling this `_validate_uri_path` method is called several times, especially in the `cloud_get` / `cloud_post` etc. methods, mainly with the `uri` argument being parts so a `list`,

Though, at a specific place it seems to be an URL, so I presume it'd have the scheme separator :

https://github.com/opengisch/qfieldsync/blob/ccab331b14bb40738265393bc6d54cb3d634864d/qfieldsync/core/cloud_api.py#L627-L629

qfieldsync/core/cloud_api.py
```diff
@@ -936,10 +936,38 @@ def cloud_upload_files(
 
         return reply
 
+    def _validate_uri_path(self, uri: Union[str, list[str]]) -> None:
+        """Validate that incoming URI components do not contain dangerous path elements."""
```
suricactus:
```suggestion
    def _clean_uri_path(self, uri: str) -> str:
        """Clean the URI components to not contain dangerous path elements, e.g. double slashes."""
```

Can we change the function instead of throwing an error, to just clean the bad parts of the URL, e.g. if I pass: `https://example.com/hello///world//` -> `https://example.com/hello/world/` ?

Also, I see you also allowed `uri: list[str]`, is it safe to switch to `str` inputs only?

qfieldsync/core/cloud_api.py
SeqLaz: @gounux you are completely correct!.... Many thanks for spot it! :ninja:

SeqLaz: Managed in the last commit!

qfieldsync/core/cloud_api.py
SeqLaz: Yup, indeed much better to only clean the URI, for this I used a Regex for cleaning up, also this way in the future we can add more cleaning practice

SeqLaz:
@suricactus, @gounux Thanks for the suggestion, they were addressed, so I squashed to only have the suggestions

qfieldsync/core/cloud_api.py
```diff
@@ -936,24 +936,32 @@ def cloud_upload_files(
 
         return reply
 
+    def _clean_uri_path(self, uri: str) -> str:
+        """Clean the URI string to remove problematic URIs with double slash"""
+        return re.sub(r"(?<!:)/{2,}", "/", uri)
```
suricactus: This is a bit naive now, but should be fine for now?

I mean if you have `https://example.com/path://to/something` -> `https://example.com/path://to/something` (no change), right?

SeqLaz: @suricactus What if I use the parse to get the protocol, network, and then just clean the path?

SeqLaz: With the last commit I addressed to only clean the path on the URI

qfieldsync/core/cloud_api.py
```diff
@@ -936,24 +936,44 @@ def cloud_upload_files(
 
         return reply
 
+    def _clean_uri_path(self, uri: str) -> str:
+        """Clean the URI string to remove problematic URIs with double slash"""
+        parsed = urllib.parse.urlparse(uri)
+        cleaned_path = re.sub(r"/{2,}", "/", parsed.path)
```
suricactus:
```suggestion
        """Clean the URI string to remove problematic filepath elements, such as double slashes, two dots etc."""
        parsed = urllib.parse.urlparse(uri)
        cleaned_path = "/" + parsed.path
        cleaned_path = cleaned_path.replace("/../", "/")
        cleaned_path = cleaned_path.replace("/./", "/")
        cleaned_path = re.sub(r"/{2,}", "/", cleaned_path)
```

Normalize path even further. Note that the proposed change enforces an opening slash. I think it should not be a problem, but please double check.

SeqLaz: @suricactus  Do you think would be better to handle it with `posixpath.normpath()`, as there could also be some errors or missed problematic paths coming from Windows?

SeqLaz: With something like this:

```python
    def _clean_uri_path(self, uri: str) -> str:
        """Clean the URI string to remove problematic filepath elements, such as double slashes, two dots etc"""
        parsed = urllib.parse.urlparse(uri)

        path = re.sub(r"/{2,}", "/", parsed.path)

        normalized_path = posixpath.normpath(path)

        if (parsed.path.endswith("/") or uri.endswith("/")) and not normalized_path.endswith("/"):
            normalized_path += "/"

        unparsed_url_str = urllib.parse.urlunparse(
            (
                parsed.scheme,
                parsed.netloc,
                normalized_path,
                parsed.params,
                parsed.query,
                parsed.fragment,
            )
        )

        return unparsed_url_str
```

suricactus: Do we still need the substitute, I think `normpath` is going to solve it for us?

As for the `normalized_path` trailing slash, I think you are missing braces. In any case, I think you should check only for `parsed.path.endswith("/")`.

We can also rename the function to `_normalize_uri`, I think it better describes what it does.

SeqLaz: > Do we still need the substitute, I think `normpath` is going to solve it for us?

No, as the `normpath` automatically remove the double slashed (confirmed with tests)

> As for the `normalized_path` trailing slash, I think you are missing braces. In any case, I think you should check only for `parsed.path.endswith("/")`.

Great!

> We can also rename the function to `_normalize_uri`, I think it better describes what it does.

Indeed more readable on what really doing

Suggestions addressed in the last commit!.... Thanks @suricactus! :ninja:

suricactus (approved):
Thanks!

PR #819, opened 2026-07-20
Support embedding the photo stamp logo in the project

kaustuvpokharel:
Right now the stamp decoration image has to be a file sitting next to the project, which is awkward to keep track of when sharing or packaging. This swaps the plain file picker for QgsImageSourceLineEdit, so you can embed the image straight into the project instead, no separate file to carry around, planned by @nirvn

Screenshot 2026-07-19 at 11 11 58 PM

### DIFF
--- a/qfieldsync/ui/image_stamping_configuration_widget.ui
+++ b/qfieldsync/ui/image_stamping_configuration_widget.ui
@@ -58,7 +58,7 @@
        </widget>
       </item>
       <item row="2" column="1">
-       <widget class="QgsFileWidget" name="customImageDecorationFile" native="true"/>
+       <widget class="QgsImageSourceLineEdit" name="customImageDecorationFile"/>
       </item>
       <item row="3" column="0" colspan="2">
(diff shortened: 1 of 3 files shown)


nirvn (dismissed):

qfieldsync/ui/image_stamping_configuration_widget.ui
```diff
@@ -170,9 +170,9 @@
    <container>1</container>
   </customwidget>
   <customwidget>
-   <class>QgsFileWidget</class>
+   <class>QgsImageSourceLineEdit</class>
    <extends>QWidget</extends>
-   <header>qgsfilewidget.h</header>
+   <header>qgis.gui</header>
```
nirvn, line 175:
```suggestion
   <header>qgsfilecontentsourcelineedit.h</header>
```

nirvn: Well, after some testing with Kaustuv, the lesson learned here is, a .ui file being loaded in Python won't like a <header> value that's not a direct equivalent to its class name.

PR #825, opened 2026-08-11
Add the option to create/switch to `TEMPLATE` or `REGULAR` projects

manylon:
In the past we only had one type of projects in QFieldCloud - "REGULAR". With those projects we can do anything - sync from QField, QGIS, CLI etc.

However, https://github.com/opengisch/QFieldCloud/pull/1735  introduced a new `TEMPLATE` project type.

This PR introduces the ability to create a new project and mark it as `TEMPLATE`, or change the type of existing project between `TEMPLATE` and `REGULAR`.

- Add a `ProjectType` enum (`regular`/`template`/`shared_datasets`) and a `CloudProject.project_type` property, mirroring QFieldCloud's new `project_type` field
- Let users pick `Regular`/`Template` when creating a new project
- Let users switch between `Regular`/`Template` when editing an existing project
- Show a disabled dropdown reading "Shared datasets", for projects whose type is `shared_datasets`
- Fall back to "regular" when talking to older API servers that don't send the `project_type` field yet

### DIFF
--- a/qfieldsync/core/cloud_api.py
+++ b/qfieldsync/core/cloud_api.py
@@ -56,7 +56,7 @@ from qgis.PyQt.QtNetwork import (
     QNetworkRequest,
 )
 
-from qfieldsync.core.cloud_project import CloudProject
+from qfieldsync.core.cloud_project import CloudProject, ProjectType
 from qfieldsync.core.preferences import Preferences
 from qfieldsync.utils.qt_utils import strip_html
 
@@ -583,7 +583,12 @@ class CloudNetworkAccessManager(QObject):
         return response.json()
 
     def create_project(
-        self, name: str, owner: str, description: str, is_public: bool
+        self,
+        name: str,
+        owner: str,
+        description: str,
+        is_public: bool,
+        project_type: ProjectType = ProjectType.REGULAR,
     ) -> QNetworkReply:
         """Create a new QFieldCloud project"""
         return self.cloud_post(
@@ -593,6 +598,7 @@ class CloudNetworkAccessManager(QObject):
                 "owner": owner,
                 "description": description,
                 "is_public": is_public,
+                "project_type": project_type.value,
             },
         )
 
--- a/qfieldsync/core/cloud_project.py
+++ b/qfieldsync/core/cloud_project.py
@@ -21,7 +21,7 @@
 import sqlite3
 from collections.abc import Iterator
 from datetime import datetime, timezone
-from enum import IntFlag
+from enum import Enum, IntFlag
 from pathlib import Path
 from typing import Any, Optional
 
@@ -44,6 +44,12 @@ class ProjectFileCheckout(IntFlag):
     Local = 1
     Cloud = 2
     LocalAndCloud = 3
+
+
+class ProjectType(str, Enum):
+    REGULAR = "regular"
+    TEMPLATE = "template"
+    SHARED_DATASETS = "shared_datasets"
 
 
 class ProjectFile:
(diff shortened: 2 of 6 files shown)


duke-nyuki:
Task linked: [QF-8625 Add the option to create/switch to TEMPLATE or REGULAR projects in QFieldSync](https://app.clickup.com/t/2192114/QF-8625)

qfieldsync/core/cloud_api.py
```diff
@@ -583,7 +583,12 @@ def get_projects_not_async(self, should_include_public: bool = False) -> list[di
         return response.json()
 
     def create_project(
-        self, name: str, owner: str, description: str, is_public: bool
+        self,
+        name: str,
+        owner: str,
+        description: str,
+        is_public: bool,
+        project_type: str = ProjectType.Regular.value,
```
suricactus:
```suggestion
        project_type: ProjectType = ProjectType.Regular,
```

Always prefer enums over their values.

qfieldsync/core/cloud_api.py
```diff
@@ -593,6 +598,7 @@ def create_project(
                 "owner": owner,
                 "description": description,
                 "is_public": is_public,
+                "project_type": project_type,
```
suricactus:
```suggestion
                "project_type": project_type.value,
```

Always prefer enums over their values and convert them to a value at the very end where it is needed.

qfieldsync/core/cloud_project.py
```diff
@@ -255,6 +261,11 @@ def description(self) -> str:
     def is_public(self) -> bool:
         return self._data["is_public"]
 
+    @property
+    def project_type(self) -> str:
+        # TODO @manylon: remove default fallback once all API servers support the `project_type` key
+        return self._data.get("project_type", ProjectType.Regular.value)
```
suricactus:
```suggestion
    @property
    def project_type(self) -> ProjectType:
        # Servers APIs older than QFC v26.XX do not have the `project_type` property, fallback to support older QFC API versions
        try:
            return ProjectType(self._data.get("project_type"))
        except WhateverExceptionHere as err:
            if self._data.get("is_shared_datasets", False):
                return ProjectType.SHARED_DATASETS

        return ProjectType.Regular
```

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -841,6 +846,29 @@ def show_project_form(self) -> None:
 
         self.projectOwnerLineEdit.setText(self.current_cloud_project.owner)
 
+        if self.current_cloud_project.project_type == ProjectType.SharedDatasets.value:
+            self.projectTypeComboBox.clear()
+            self.projectTypeComboBox.addItem(self.tr("Shared datasets"))
```
suricactus:
```suggestion
            self.projectTypeComboBox.addItem(self.tr("Shared datasets"), ProjectType.SharedDatasets.value)
```

suricactus (commented):
Nice first contribution. A few small comments.

qfieldsync/gui/cloud_projects_dialog.py
```diff
@@ -841,6 +846,31 @@ def show_project_form(self) -> None:
 
         self.projectOwnerLineEdit.setText(self.current_cloud_project.owner)
 
+        if self.current_cloud_project.project_type == ProjectType.SHARED_DATASETS.value:
+            self.projectTypeComboBox.clear()
+            self.projectTypeComboBox.addItem(
+                self.tr("Shared datasets"), ProjectType.SHARED_DATASETS.value
```
suricactus: I assume you tried setting directly the enum as value, instead of the value of the enum:

```suggestion
                self.tr("Shared datasets"), ProjectType.SHARED_DATASETS
```

I had some hopes that QVariant will handle these easily without the need to explicitly convert. https://doc.qt.io/qt-6/qcombobox.html#addItem

suricactus (approved):
Thanks!

manylon:
Thanks too, will clean history and merge.

PR #829, opened 2026-08-29
Improve cloud project deletion by avoiding overly wide dialog and allowing text selection

nirvn:
Deleting cloud projects in QFieldSync is _painful_ when compared with deletion on app.qfield.cloud.

The main reason is that for long user/project names, the (good) requirement to confirm via project name becomes a real PITA especially when you have to delete multiple projects in a row.

On the web interface, users can copy/paste the name from the HTML confirmation dialog. Let's allow that in QFieldSync too. It'll still require users to provide a confirmation, but they'll be able to fasten the process via copy/paste.

The confirmation dialog was also insanely wide for long user/project names, this PR fixes that too.

Result:

Screenshot From 2026-08-29 11-20-27

### DIFF
--- a/qfieldsync/gui/cloud_projects_dialog.py
+++ b/qfieldsync/gui/cloud_projects_dialog.py
@@ -51,6 +51,7 @@ from qgis.PyQt.QtWidgets import (
     QHBoxLayout,
     QHeaderView,
     QInputDialog,
+    QLabel,
     QMenu,
     QMessageBox,
     QPushButton,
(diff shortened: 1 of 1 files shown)


suricactus:
Would you mind providing also a "before" screenshot. I couldn't imagine what it used to before the changes.

Anyways,  thanks!

suricactus (approved):

nirvn:
@suricactus , here we go:

image

PR #830, opened 2026-08-29
Add bandit security check to avoid blocked release when publishing updates

nirvn:
I'd rather have us be made aware during development than having a blocked release :)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -18,14 +18,17 @@ jobs:
     name: Code check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v7
+      - name: Checkout Code
+        uses: actions/checkout@v7
 
-    - uses: actions/setup-python@v7
+      - name: Setup Python
+        uses: actions/setup-python@v7
 
-    - uses: pre-commit/action@v3.0.1
-      env:
-        # set the `no-commit-to-branch` to be skipped, otherwise the check fails when we merge to `master` branch
-        SKIP: 'no-commit-to-branch'
+      - name: Pre-Commit
+        uses: pre-commit/action@v3.0.1
+        env:
+          # set the `no-commit-to-branch` to be skipped, otherwise the check fails when we merge to `master` branch
+          SKIP: 'no-commit-to-branch'
 
   pyqt6-check:
     name: PyQt6 compatibility check
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,3 +46,12 @@ repos:
     rev: v2.1.0
     hooks:
     -   id: dead
+
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.9.4
+    hooks:
+        # Critical mandatory checks from QGIS plugin repository that prevent publishing a new version of the plugin
+        # NOTE: Ruff uses `S` and Bandit `B` prefixes, so both codes must be excluded per line.
+        # Keep in sync with https://github.com/qgis/QGIS-Plugins-Website/blob/master/qgis-app/plugins/management/commands/data/bandit_rules.json
+      - id: bandit
+        args: ["-t", "B102,B103,B105,B106,B107,B201,B202,B301,B302,B304,B305,B306,B307,B312,B321,B323,B401,B402,B412,B413,B501,B502,B503,B505,B506,B507,B601,B602,B604,B605,B609,B610,B611,B612,B613,B701", "-r", "."]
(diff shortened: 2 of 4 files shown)


gounux:
@nirvn, I guess that there is a bandit pre-commit hook that can be added and configured in the `.pre-commit-config.yml` file, if this can help spot potential issues locally,

m-kuhn:
I have this link in my clipboard while reading the comment by @gounux .pre-commit-hooks.yaml
https://github.com/PyCQA/bandit/blob/main/.pre-commit-hooks.yaml

suricactus:
I think most if not all bandit checks are part of ruff, we need to enable them, rather than introducing a new tool.

nirvn:
@suricactus , it's not only about the check itself, it's about making sure we add the relevant bandit whitelist tags (e.g.: https://github.com/opengisch/qfieldsync/pull/830/changes#diff-b06b19d86274a315bf36c35bc43b352c8c87ff775e1d661b48de1fdb210aaff9R70)

suricactus:
Ah, that's unfortunate that bandit and ruff use different prefix (B vs S). IMO this is an upstream issue, but this PR is a good move for now. Thanks!

nirvn:
@gounux , @m-kuhn , pre-commit hook added (good call).

nirvn:
@suricactus , +1 ? If so, let's merge this and release 4.23.1.

gounux:
@nirvn thanks for the pre-commit hook :)  
Since there is already a pre-commit check in the CI - that would then also check bandit hooks, wondering what'd be the benefits of having a new workflow / CI run ?

nirvn:
@gounux , the ci run integrates with GitHub code scanner. So in theory for ppl who don't have pre-commit installed locally that'd integrate errors within the PR page itself

suricactus:
If one wants to contribute to QFieldSync, they must have precommit locally.

.github/workflows/continuous_integration.yml
```diff
@@ -18,14 +18,33 @@ jobs:
...
-      env:
-        # set the `no-commit-to-branch` to be skipped, otherwise the check fails when we merge to `master` branch
-        SKIP: 'no-commit-to-branch'
+    steps:
+      - name: Bandit Scan
+        uses: PyCQA/bandit-action@v1
+        with:
+          # critical checks from https://github.com/qgis/QGIS-Plugins-Website/blob/master/qgis-app/plugins/management/commands/data/bandit_rules.json
+          tests: B102,B103,B105,B106,B107,B201,B202,B301,B302,B304,B305,B306,B307,B312,B321,B323,B401,B402,B412,B413,B501,B502,B503,B505,B506,B507,B601,B602,B604,B605,B609,B610,B611,B612,B613,B701
+          exit_zero: false
```
suricactus: I find this as redundant `bandit` job redundant, we already have `pre-commit` running in the `code_check` job, which is telling us what is wrong in the job logs. In the meantime the change in `.pre-commit-config.yaml` guarantees the checks will run both locally and in the CI.

nirvn:
@gounux , @suricactus , the majority wins, the extra workflow has been taken down.

.pre-commit-config.yaml
```diff
@@ -46,3 +46,9 @@ repos:
     rev: v2.1.0
     hooks:
     -   id: dead
+
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.9.4
+    hooks:
+      - id: bandit
+        args: ["-t", "B102,B103,B105,B106,B107,B201,B202,B301,B302,B304,B305,B306,B307,B312,B321,B323,B401,B402,B412,B413,B501,B502,B503,B505,B506,B507,B601,B602,B604,B605,B609,B610,B611,B612,B613,B701", "-r", "."]
```
suricactus, line 57:
```suggestion
          # Critical mandatory checks from QGIS plugin repository that prevent publishing a new version of the plugin
          # NOTE: Ruff uses `S` and Bandit `B` prefixes, so both codes must be excluded per line.
          # Keep in sync with https://github.com/qgis/QGIS-Plugins-Website/blob/master/qgis-app/plugins/management/commands/data/bandit_rules.json
        args: ["-t", "B102,B103,B105,B106,B107,B201,B202,B301,B302,B304,B305,B306,B307,B312,B321,B323,B401,B402,B412,B413,B501,B502,B503,B505,B506,B507,B601,B602,B604,B605,B609,B610,B611,B612,B613,B701", "-r", "."]
```

suricactus (commented):
One last small comment to add a comment about the unfortunate situation we are in. Thanks! 

While the `libqfieldsync` codebase does not go trough the same checks as it's shipped within a whl file, I think it's worth it to add the very same change there.

And probably the `xlsformconverter` and `ordered-select` plugin too.

nirvn:
@suricactus , yeah, once we settle on this PR, I was planning to do it for these repos and add a note internally to suggest for our other plugin maintainers to do the same.

Your suggested comment has been added

nirvn:
@suricactus , was your heart an approval? :)

suricactus (approved):
💚

